PR #8, opened 2018-05-17
Added prototype for profile pages

vocatwk:
Most of the new code is in ProfileServlet.java. I have implemented only things I need to get the doGet() method working and I have a method header for the doPost(). I have added this servlet class and its corresponding url pattern in web.xml. I have also added a profile.jsp file that has the same nav bar as the other pages but with an additional message. (I've attached a screenshot below). I have also edited all jsp files(just the nav bars to include a link to the profile of the user logged in.) I've also added a method in MessageStore.java that lets me access messages in any conversation from a specific author(this is really for functions I have yet to implement so it's not used for prototype)
screen shot 2018-05-17 at 5 17 21 pm

### DIFF
--- a/src/main/java/codeu/controller/ProfileServlet.java
+++ b/src/main/java/codeu/controller/ProfileServlet.java
@@ -61,7 +61,7 @@ public class ProfileServlet extends HttpServlet {
 
     if(requestUrl.length() <= "/profile/".length()){
       // if user navigates to "/profile/" without a specific user
-      request.getRequestDispatcher("WEB-INF/view/index.jsp").forward(request,response);
+      request.getRequestDispatcher("/index.jsp").forward(request,response);
       return;
     }
     String subjectName = requestUrl.substring("/profile/".length());
@@ -71,7 +71,7 @@ public class ProfileServlet extends HttpServlet {
 
     if(subject == null) {
       // couldn't file profile, redirect to index.jsp
-      request.getRequestDispatcher("/WEB-INF/view/index.jsp").forward(request, response);
+      request.getRequestDispatcher("/index.jsp").forward(request, response);
       return;
     }
 
--- a/src/main/java/codeu/controller/ProfileServlet.java
+++ b/src/main/java/codeu/controller/ProfileServlet.java
@@ -1,0 +1,97 @@
+package codeu.controller;
+
+import codeu.model.data.Conversation;
+import codeu.model.data.Message;
+import codeu.model.data.User;
+import codeu.model.store.basic.ConversationStore;
+import codeu.model.store.basic.MessageStore;
+import codeu.model.store.basic.UserStore;
+import java.io.IOException;
+import java.time.Instant;
+import java.util.List;
+import java.util.UUID;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.jsoup.Jsoup;
+import org.jsoup.safety.Whitelist;
+
+/** Servlet class responsible for the profile page. */
+public class ProfileServlet extends HttpServlet {
+   
+  private MessageStore messageStore;
+
+  /** Store class that gives access to Users. */
+  private UserStore userStore;
+
+  /** Set up state for handling profile requests. */
+  @Override
+  public void init() throws ServletException {
+    super.init();
+    setMessageStore(MessageStore.getInstance());
+    setUserStore(UserStore.getInstance());
+  }
+
+  /**
+   * Sets the MessageStore used by this servlet. This function provides a common setup method for
+   * use by the test framework or the servlet's init() function.
+   */
+  void setMessageStore(MessageStore messageStore) {
+    this.messageStore = messageStore;
+  }
+
+  /**
+   * Sets the UserStore used by this servlet. This function provides a common setup method for use
+   * by the test framework or the servlet's init() function.
+   */
+  void setUserStore(UserStore userStore) {
+    this.userStore = userStore;
+  }
+
+  /**
+   * This function fires when a user navigates to the profile page. It gets the profile name from
+   * the URL, finds the corresponding profile subject, and fetches the messages sent by that subject.
+   * It then forwards to profile.jsp for rendering.
+   */
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response)
+      throws IOException, ServletException {
+    String requestUrl = request.getRequestURI();
+
+    if(requestUrl.length() <= "/profile/".length()){
+      // if user navigates to "/profile/" without a specific user
+      request.getRequestDispatcher("WEB-INF/view/index.jsp").forward(request,response);
+      return;
+    }
+    String subjectName = requestUrl.substring("/profile/".length());
+	
+    
+    User subject = userStore.getUser(subjectName);
+
+    if(subject == null) {
+      // couldn't file profile, redirect to index.jsp
+      request.getRequestDispatcher("/WEB-INF/view/index.jsp").forward(request, response);
+      return;
+    }
+
+    request.setAttribute("subject", subjectName);
+    UUID subjectId = subject.getId();
+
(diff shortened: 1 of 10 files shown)


yebrahim:
Nice!
@vocatwk can you fix the PR's title to summarize what it does?

@jasminechau2 and @lealgamb, can you guys take a look whenever you have the chance?

vocatwk:
Oops sorry about that. Fixed!

lealgamb:
This looks good to me. Breaking down the changes into different commits and testing the changes locally was definitely helpful!

src/main/java/codeu/controller/ProfileServlet.java
```diff
@@ -0,0 +1,97 @@
+package codeu.controller;
+
+import codeu.model.data.Conversation;
```
yebrahim: Are all these imports needed here? It seems to me some of these aren't used in this file.

src/main/java/codeu/controller/ProfileServlet.java
```diff
@@ -0,0 +1,97 @@
...
+   * This function fires when a user navigates to the profile page. It gets the profile name from
+   * the URL, finds the corresponding profile subject, and fetches the messages sent by that subject.
+   * It then forwards to profile.jsp for rendering.
+   */
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response)
+      throws IOException, ServletException {
+    String requestUrl = request.getRequestURI();
+
+    if(requestUrl.length() <= "/profile/".length()){
```
yebrahim: It's great that you've thought about these edge cases.
For this case, it seems to me you want to redirect the browser to the index page, instead of serving it with the same url? Something like `response.sendRedirect("/");`?

src/main/java/codeu/controller/ProfileServlet.java
```diff
@@ -0,0 +1,97 @@
...
+      // if user navigates to "/profile/" without a specific user
+      request.getRequestDispatcher("/index.jsp").forward(request,response);
+      return;
+    }
+    String subjectName = requestUrl.substring("/profile/".length());
+	
+    
+    User subject = userStore.getUser(subjectName);
+
+    if(subject == null) {
```
yebrahim: Again, it's great you're covering this. For now, we can also redirect to '/' like I suggested above?
For later, how about showing an error page, like a 404 or something, when the user navigates to a page that doesn't exist? Just something to keep in mind.

src/main/java/codeu/controller/ProfileServlet.java
```diff
@@ -0,0 +1,97 @@
...
+  public void doGet(HttpServletRequest request, HttpServletResponse response)
+      throws IOException, ServletException {
+    String requestUrl = request.getRequestURI();
+
+    if(requestUrl.length() <= "/profile/".length()){
+      // if user navigates to "/profile/" without a specific user
+      request.getRequestDispatcher("/index.jsp").forward(request,response);
+      return;
+    }
+    String subjectName = requestUrl.substring("/profile/".length());
```
yebrahim: There's another way you've used below to get the user name, using the `request.getSession()` object. Do you see the difference?

src/main/java/codeu/model/store/basic/MessageStore.java
```diff
@@ -84,6 +84,19 @@ public void addMessage(Message message) {
...
     return messagesInConversation;
   }
+  /** Access the current set of Messages from the given author. */
+  public List<Message> getMessagesFromAuthor(UUID authorId) {
+
+    List<Message> messagesFromAuthor = new ArrayList<>();
+    
+    for(Message message : messages) {
+      if(message.getAuthorId().equals(authorId)) {
+	messagesFromAuthor.add(message);
```
yebrahim: Seems like there's a little indentation issue here. This line uses tabs instead of spaces? :)

src/main/webapp/WEB-INF/view/chat.jsp
```diff
@@ -50,7 +50,8 @@ List<Message> messages = (List<Message>) request.getAttribute("messages");
     <a id="navTitle" href="/">CodeU Chat App</a>
     <a href="/conversations">Conversations</a>
       <% if (request.getSession().getAttribute("user") != null) { %>
-    <a>Hello <%= request.getSession().getAttribute("user") %>!</a>
+    <a href="/profile/<%= request.getSession().getAttribute("user") %>">
```
yebrahim: nitpick: how about saving the username into a variable at the top of the page so it can be reused?

src/main/webapp/WEB-INF/view/profile.jsp
```diff
@@ -0,0 +1,39 @@
+<%@ page import="java.util.List" %>
+<%@ page import="codeu.model.data.Conversation" %>
+<%@ page import="codeu.model.data.Message" %>
+<%@ page import="codeu.model.store.basic.UserStore" %>
+<%
+List<Message> messages = (List<Message>) request.getAttribute("messages");
```
yebrahim: Do we need this line?

src/main/webapp/WEB-INF/view/profile.jsp
```diff
@@ -0,0 +1,39 @@
+<%@ page import="java.util.List" %>
```
yebrahim: Do we need all these imports?

src/main/java/codeu/controller/ProfileServlet.java
vocatwk: Oh, I see. Hadn't understood the difference before. I will change it.

src/main/java/codeu/controller/ProfileServlet.java
vocatwk: Yeah, you're right. I will make sure to only keep the ones needed.

src/main/java/codeu/controller/ProfileServlet.java
vocatwk: hmm .. I'm not sure what you mean by this. I was able to get the username from the session because it was previously saved but I have yet to save the name of the user whose profile we're navigating to, right? Am I misunderstanding something?

src/main/java/codeu/controller/ProfileServlet.java
vocatwk: 404 sounds like a good idea. Thanks!

src/main/java/codeu/model/store/basic/MessageStore.java
vocatwk: will fix.

src/main/webapp/WEB-INF/view/chat.jsp
vocatwk: I had actually thought about that too. Don't know why I didn't end up doing. I will correct it though.

src/main/webapp/WEB-INF/view/profile.jsp
vocatwk: I added this for later use because we're going to be iterating through the messages sent by the user whose profile we're viewing but you're right, this isn't needed for the prototype.

src/main/java/codeu/controller/ProfileServlet.java
yebrahim: I guess it depends what we're trying to do here. Do we want users to have access to each others profiles without logging in, or do they have to login first?

src/main/java/codeu/controller/ProfileServlet.java
yebrahim: Cool. I wouldn't worry about it though, I'd open an issue and forget about it for now.

src/main/webapp/WEB-INF/view/profile.jsp
yebrahim: Cool, it's fine to keep this one. In general, it's good to keep the code scoped to the change though. If we need to revert a change for any reason in the future, it should only revert the functionality it added, nothing more or less. :)

src/main/java/codeu/controller/ProfileServlet.java
vocatwk: I wondered the same thing too but since they can see conversation topics and chats without logging in, I figured they could see profiles as well.

PR #10, opened 2018-05-18
Added Prototype for Activity Feed

lealgamb:
I created the activity feed servlet and it's corresponding jsp file. I also updated web.xml to map to the corresponding url. Currently, the activity feed page doesn't fetch any data and only has the doGet() method implemented. Here's a screenshot:

image

### DIFF
--- a/src/main/java/codeu/controller/ActivityFeedServlet.java
+++ b/src/main/java/codeu/controller/ActivityFeedServlet.java
@@ -1,0 +1,41 @@
+package codeu.controller;
+
+import java.io.IOException;
+import java.time.Instant;
+import java.util.List;
+import java.util.UUID;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.jsoup.Jsoup;
+import org.jsoup.safety.Whitelist;
+
+/** Servlet class responsible for the activty feed page. */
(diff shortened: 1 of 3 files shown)


src/main/webapp/WEB-INF/view/activityfeed.jsp
```diff
@@ -0,0 +1,28 @@
...
+      <a href="/profile/<%=request.getSession().getAttribute("user") %>">
+         Hello <%= request.getSession().getAttribute("user") %>!</a>
+    <% } else{ %>
+      <a href="/login">Login</a>
+    <% } %>
+    <a href="/about.jsp">About</a>
+  </nav>
+
+  <div id="container">
+  	<h1>Activity</h1>
```
yebrahim: nitpick: tabs vs spaces ;)

src/main/java/codeu/controller/ActivityFeedServlet.java
```diff
@@ -0,0 +1,41 @@
+package codeu.controller;
+
+import java.io.IOException;
```
yebrahim: As @vocatwk mentioned, I'd go over these to make sure nothing is being imported unnecessarily.

jasminechau2:
It looks good! Keep on working on it!

src/main/java/codeu/controller/ActivityFeedServlet.java
lealgamb: Will do, thank you!

PR #23, opened 2018-05-23
Fetched conversations for activity feed

lealgamb:
I integrated the conversations into the activity feed. I also combined events by using a sorted map so that I could easily sort the events chronologically by their creation time. I'll add the messages and update the tests later on.

Here's a screenshot:
image

### DIFF
--- a/src/main/java/codeu/controller/ActivityFeedServlet.java
+++ b/src/main/java/codeu/controller/ActivityFeedServlet.java
@@ -1,9 +1,16 @@
 package codeu.controller;
 
 import codeu.model.data.User;
+import codeu.model.data.Conversation;
 import codeu.model.store.basic.UserStore;
+import codeu.model.store.basic.ConversationStore;
 import java.io.IOException;
+import java.time.Instant;
 import java.util.List;
+import java.util.Comparator;
+import java.util.Map;
+import java.util.HashMap;
+import java.util.TreeMap;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
@@ -15,19 +22,31 @@ public class ActivityFeedServlet extends HttpServlet {
   /** Store class that gives access to Users. */
   private UserStore userStore;
 
+  /** Store class that gives access to Conversations. */
+  private ConversationStore conversationStore;
+
   /** Set up state for handling activty feed requests. */
   @Override
   public void init() throws ServletException {
   	super.init();
     setUserStore(UserStore.getInstance());
+    setConversationStore(ConversationStore.getInstance());
   }
 
   /**
-   * Sets the UserStore used by this servlet. This function provides a common setup method for use
-   * by the test framework or the servlet's init() function.
+   * Sets the UserStore used by this servlet. This function provides a common 
+   * setup method for use by the test framework or the servlet's init() function.
    */
   void setUserStore(UserStore userStore) {
     this.userStore = userStore;
+  }
+
+  /**
+   * Sets the ConversationStore used by this servlet. This function provides a 
+   * common setup method for use by the test framework or the servlet's init() function.
+   */
+  void setConversationStore(ConversationStore conversationStore) {
+    this.conversationStore = conversationStore;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/main/java/codeu/controller/ActivityFeedServlet.java
```diff
@@ -15,30 +22,66 @@
...
       throws IOException, ServletException {
+    // Get and set Users
     List<User> users = userStore.getAllUsers();
     request.setAttribute("users", users);
+
+    // Get and set Conversations
+    List<Conversation> conversations = conversationStore.getAllConversations();
+    request.setAttribute("conversations", conversations);
+
+    // Make, sort, and set the event map
```
yebrahim: If we care about having conversations in order, I wonder if we want to change the way they're stored. Think about having *a lot* of conversations in the database, are we going to sort them all on every request?

This is an exploratory question though, it's fine if we start with a simple implementation for now and then iterate to make it better.

src/main/java/codeu/controller/ActivityFeedServlet.java
```diff
@@ -15,30 +22,66 @@
...
+    Map<Instant, Object> unsortedEventsMap = new HashMap<Instant, Object>();
+    for(User user : users)
+      unsortedEventsMap.put(user.getCreationTime(), user);
+    for(Conversation conversation : conversations)
+      unsortedEventsMap.put(conversation.getCreationTime(), conversation);
+    request.setAttribute("unsortedEventsMap", unsortedEventsMap);
```
yebrahim: Does seem that you're actually using this anywhere?

src/main/java/codeu/controller/ActivityFeedServlet.java
```diff
@@ -15,30 +22,66 @@
...
+    // Sort the event map by key in descending order
+    Map<Instant, Object> sortedEventsMap = 
+      new TreeMap<Instant, Object>(new Comparator<Instant>() {
+        @Override
+        public int compare(Instant o1, Instant o2) {
+          return o2.compareTo(o1);
+        }
+      });
+    sortedEventsMap.putAll(unsortedEventsMap);
+    request.setAttribute("sortedEventsMap", sortedEventsMap);
```
yebrahim: If the above comment is correct and all we need is the sorted map, how about we create it and insert items into it directly?

src/main/webapp/WEB-INF/view/activityfeed.jsp
```diff
@@ -28,17 +38,40 @@
...
+          %>
+            <a href="/profile/<%= user.getName() %>"><%= user.getName() %></a> 
+              joined!
+          </li>
+          <%
+          }
+          else if (entry.getValue() instanceof Conversation) {
+            Conversation conversation = (Conversation) entry.getValue();
+            User conversationOwner = null;
+            for (User user : users) {
```
yebrahim: If we have the user id, can we get the user from the UserStore directly instead of searching for it linearly in the list of users?

src/main/webapp/WEB-INF/view/activityfeed.jsp
```diff
@@ -28,17 +38,40 @@
...
+        for(Map.Entry<Instant, Object> entry : sortedEventsMap.entrySet()) {
+          Date date = Date.from(entry.getKey());
         %>
-        <li><b><%= date %>: </b> 
-          <a href="/profile/<%= user.getName() %>">
-          <%= user.getName() %></a> joined!</li>
+          <li>
+            <b><%= date %>:</b>
+          <%
+          if(entry.getValue() instanceof User) {
```
yebrahim: An alternative implementation to consider is to build a new class, let's call it `Event`, that contains details such as the type of event, date, the username if it's a join event, the conversation id and author if it's a new conversation.. etc, then passing it to the jsp, which uses it directly to render the HTML.

src/main/java/codeu/controller/ActivityFeedServlet.java
```diff
@@ -15,30 +22,66 @@
   /** Store class that gives access to Users. */
   private UserStore userStore;
 
+  /** Store class that gives access to Conversations. */
```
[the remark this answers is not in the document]
lealgamb: Will do.

src/main/java/codeu/controller/ActivityFeedServlet.java
lealgamb: I was thinking of sorting them as they're created, but then when I combined the conversations with users I didn't go through with that. I'll look into it with the Event class you mentioned below.

src/main/java/codeu/controller/ActivityFeedServlet.java
lealgamb: You're right, I'll remove it.

src/main/java/codeu/controller/ActivityFeedServlet.java
lealgamb: Will do.

src/main/webapp/WEB-INF/view/activityfeed.jsp
lealgamb: I wanted to do this but was unsure how to go about it. Would I create an instance of UserStore in the servlet class and set it as an attribute to the request, or create the instance in the jsp file?

src/main/webapp/WEB-INF/view/activityfeed.jsp
lealgamb: I'll try this implementation, it will be a lot cleaner. Thanks!

src/main/webapp/WEB-INF/view/activityfeed.jsp
yebrahim: I haven't tried this so I don't know if it's doable, but can we pass the reference to the `UserStore` singleton to the jsp as an attribute?

lealgamb: I'll try it out asap.

src/main/java/codeu/controller/ActivityFeedServlet.java
vocatwk: you could do that initially when all the users and conversations are loaded so it doesn't do it on every request.

lealgamb: Sorry I'm not sure what you mean by this. Do you mean having it in init()? @vocatwk

src/main/java/codeu/controller/ActivityFeedServlet.java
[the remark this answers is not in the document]
lealgamb: Should it say something like "Singleton instance of [x]". Also, I used that comment based on what was already in the code base for consistency so should I still change it?

PR #25, opened 2018-05-27
Activity Feed MVP

lealgamb:
All users, conversations, and messages are now displayed in the activity feed. I added a function to access all of the messages which ideally will be removed once I make an Event class. I also removed the unnecessary unsorted map I had before. Also, I updated the CSS for this page. I still have to implement this using the Store classes so that I could access a user by their ID (same for other classes).

Here's a screenshot:
image

### DIFF
--- a/src/main/java/codeu/controller/ActivityFeedServlet.java
+++ b/src/main/java/codeu/controller/ActivityFeedServlet.java
@@ -2,14 +2,15 @@ package codeu.controller;
 
 import codeu.model.data.User;
 import codeu.model.data.Conversation;
+import codeu.model.data.Message;
 import codeu.model.store.basic.UserStore;
 import codeu.model.store.basic.ConversationStore;
+import codeu.model.store.basic.MessageStore;
 import java.io.IOException;
 import java.time.Instant;
 import java.util.List;
 import java.util.Comparator;
 import java.util.Map;
-// import java.util.HashMap;
 import java.util.TreeMap;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServlet;
@@ -25,12 +26,16 @@ public class ActivityFeedServlet extends HttpServlet {
   /** Store class that gives access to Conversations. */
   private ConversationStore conversationStore;
 
+  /** Store class that gives access to Messages. */
+  private MessageStore messageStore;
+
   /** Set up state for handling activty feed requests. */
   @Override
   public void init() throws ServletException {
   	super.init();
     setUserStore(UserStore.getInstance());
     setConversationStore(ConversationStore.getInstance());
+    setMessageStore(MessageStore.getInstance());
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/main/java/codeu/controller/ActivityFeedServlet.java
```diff
@@ -64,22 +77,24 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
     List<Conversation> conversations = conversationStore.getAllConversations();
     request.setAttribute("conversations", conversations);
 
+    // Get and set Messages
+    List<Message> messages = messageStore.getAllMessages();
+    request.setAttribute("messages", messages);
+
     // Make, sort, and set the event map
-    Map<Instant, Object> unsortedEventsMap = new HashMap<Instant, Object>();
```
yebrahim: Unrelated to this PR: you don't really need the extra `Instant` field of each record, you can just have the comparator check the object's `getCreationTime` method since they all implement it.

src/main/java/codeu/model/store/basic/MessageStore.java
```diff
@@ -65,6 +65,12 @@ private MessageStore(PersistentStorageAgent persistentStorageAgent) {
     messages = new ArrayList<>();
   }
 
+  /** Access the current set of messages known to the application. This function 
+  is temporary for the Activiy Feed MVP. */
```
yebrahim: Do we have a plan for removing this method? If so, how about adding it in a TODO here instead of the part about its being temporary?

lealgamb:
Sorry, I meant to push the updated tests after this merge.

src/main/java/codeu/model/store/basic/MessageStore.java
lealgamb: I still don't have a specific plan but I'll add a TODO since this approach would be inefficient with an Event class.

src/main/java/codeu/controller/ActivityFeedServlet.java
lealgamb: Would this mean that List<Object> would be better than the map?

yebrahim: Well you still need sorting. If you don't want to use a list and sort it, you could consider a different data structure with inherent sorting, maybe a `HashSet`?

lealgamb: Thanks, I'll look into it. Also, once the Event class is implemented would it be better to keep using a List then just traverse it in reverse? That way there's no need for sorting.

yebrahim: You'll still need to keep the list sorted on the event time.
Another alternative is to make sure things are inserted into the data store in order (which I think they are), so that you can just get the lists of convos, chats, and users, then do an efficient time-based merge, which is just the merge step of a [merge sort](https://en.wikipedia.org/wiki/Merge_sort).

lealgamb: I hadn't considered merging before! I think for now I'll traverse in reverse since they're stored in order and then figure out how to merge, thanks.

PR #33, opened 2018-05-29
Implemented Navigation Bar Changes

lealgamb:
Implemented changes based on recent issue. The code for the navigation bar is now in navbar.jsp and this file is included in all other jsp files where the previous code used to be. Also added tab for activity feed.

Here's a screenshot:

image

### DIFF
--- a/src/main/webapp/WEB-INF/view/navbar.jsp
+++ b/src/main/webapp/WEB-INF/view/navbar.jsp
@@ -1,0 +1,12 @@
+<nav>
+  <a id="navTitle" href="/">CodeU Chat App - Nemo</a>
+  <a href="/conversations">Conversations</a>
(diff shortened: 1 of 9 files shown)


src/main/webapp/WEB-INF/view/navbar.jsp
```diff
@@ -0,0 +1,12 @@
+<nav>
+  <a id="navTitle" href="/">CodeU Chat App - Nemo</a>
+  <a href="/conversations">Conversations</a>
+  <% if(request.getSession().getAttribute("user") != null){ %>
```
vocatwk: is it possible to use a variable so we don't have to keep calling these methods?

lealgamb: I'll update in the next PR.

PR #34, opened 2018-05-30
Implemented Event Class

lealgamb:
I implemented an Event class to be used for the Activity Feed. I then added getConversation() with a given Id in ConversationStore.

Also, this is my first time making a PR from another branch so let me know if there are any problems. Just wondering if I did this correctly:

```git checkout -b NewBranch```
```git pull origin OldBranch```
```git push origin NewBranch```

### DIFF
--- a/src/main/java/codeu/model/data/Event.java
+++ b/src/main/java/codeu/model/data/Event.java
@@ -1,0 +1,93 @@
+package codeu.model.data;
+
+import codeu.model.data.User;
+import codeu.model.data.Conversation;
+import codeu.model.data.Message;
+import codeu.model.store.basic.UserStore;
+import codeu.model.store.basic.ConversationStore;
+import java.time.Instant;
+import java.util.UUID;
+import java.util.List;
+import java.util.ArrayList;
+
+/**
+ * Class representing an event, which includes Users joining, Conversations 
+ * being created, and Messages being sent. 
+ */
+public class Event {
+
+  private final String type;
+  private final Instant creation;
+  /** 
+   * information stores the following as necessary:
+   * 1. Username
+   * 2. Conversation Title
+   * 3. Message Content
+   * NOTE: At minimum, must store Username
+   */
+  private final List<String> information;
+  private UserStore userStore;
+  private ConversationStore conversationStore;
+  
+  /**
+   * Constructs a new Event
+   * 
+   * @param type the type of event
+   * @param creation the time the event was created
+   * @param classType the class of the event 
+   */
+  public Event(String type, Instant creation, Object classType) {
+    this.type = type;
+    this.creation = creation;
+    information = new ArrayList<>();
+    setUserStore(userStore.getInstance());
+    setConversationStore(conversationStore.getInstance());
+
+    if (type.equals("User")) {
+      User user = (User)classType;
+      information.add(user.getName());
+    }
+    else if (type.equals("Conversation")) {
+      Conversation conversation = (Conversation)classType;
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: src/main/java/codeu/model/data/Event.java [hunk 1]
```diff
@@ -0,0 +1,93 @@
...
+    setUserStore(userStore.getInstance());
+    setConversationStore(conversationStore.getInstance());
+
+    if (type.equals("User")) {
+      User user = (User)classType;
+      information.add(user.getName());
+    }
+    else if (type.equals("Conversation")) {
+      Conversation conversation = (Conversation)classType;
+      information.add(userStore.getUser(conversation.getOwnerId()).getName());
```
vocatwk: You can also get user by Id so you don't have to call getName()

src/main/java/codeu/model/data/Event.java
```diff
@@ -0,0 +1,93 @@
...
+      information.add(conversation.getTitle());
+    }
+    else if (type.equals("Message")) {
+      Message message = (Message)classType;
+      information.add(userStore.getUser(message.getAuthorId()).getName());
```
vocatwk: same as above comment.

src/main/java/codeu/model/data/Event.java
```diff
@@ -0,0 +1,93 @@
...
+      information.add(conversationStore.getConversation(message.getConversationId()).getTitle());
```
vocatwk: You can also call getConversaiton() with id since you added that method in ConversationStore

vocatwk:
Also, about the new branch, I think it just means that you also have all the changes you had in the previous PR on this PR too. So I guess if you merge the previous first and then compare this with the master it will only have the new ones? I'm not sure if automatically updates or how that works though.

I guess you could've also pulled from master before creating this PR and then resolved conflicts after you merged your previous PR?

yebrahim:
There are a few options for the branching issue. The simplest is: if you want to create a new PR that's dependent on an outstanding PR, you can change the base branch, see this screenshot:
image

If those changes are small and you don't mind having them as part of this PR, you can keep them here and choose to rebase when merging, see the last option in the merge dropdown:
image

src/main/java/codeu/model/data/Event.java
```diff
@@ -0,0 +1,93 @@
...
+    this.conversationStore = conversationStore;
+  }
+
+  /** Returns the type of this Event. */
+  public String getType() {
+    return type;
+  }
+
+  /** Returns the creation time of this Event. */
+  public Instant getCreationTime() {
```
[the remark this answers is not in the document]
lealgamb: It's to display the creation time in the Activity Feed and possibly sort. The purpose of the class is to make the code cleaner and be able to group all events as one object.

src/main/java/codeu/model/data/Event.java
```diff
@@ -0,0 +1,93 @@
...
+  /**
+   * Constructs a new Event
+   * 
+   * @param type the type of event
+   * @param creation the time the event was created
+   * @param classType the class of the event 
+   */
+  public Event(String type, Instant creation, Object classType) {
+    this.type = type;
+    this.creation = creation;
```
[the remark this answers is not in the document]
lealgamb: It's to have access to the creation time of the actual event, like a user joining, not when the Event object is created.

lealgamb:
Thanks for the help with the branching. Just to make sure, what I did didn't create any problems, right?

yebrahim:
All looks good.
It's a good practice to create a branch per PR. Reusing branches can be tricky and you can accidentally shoot yourself in the foot.

PR #35, opened 2018-05-31
Working MVP Admin Page

jasminechau2:
I am sorry I am so late on this. This should be working I still need a way to implement who is the admin, I was thinking of a button that sends a request to current admins to approve if the user can become an admin. But you won't be able to access the page locally without being an admin. I was also wondering if there was a way to but an admin tab on the navigation bar only if the user is an admin. 
screen shot 2018-05-31 at 1 09 07 am

### DIFF
--- a/src/main/java/codeu/controller/AdminServlet.java
+++ b/src/main/java/codeu/controller/AdminServlet.java
@@ -13,6 +13,8 @@ import java.util.ArrayList;
 import java.util.List;
 import org.mindrot.jbcrypt.BCrypt;
 import java.util.UUID;
+import java.util.Arrays;
+
 
 
 /** Servlet class responsible for the Admin page. */
@@ -21,7 +23,9 @@ public class AdminServlet extends HttpServlet {
   private UserStore userStore;
   private ConversationStore conversationStore;
   private MessageStore messageStore;
-  private List<String> adminList = List.of("admin","admin1", "admin2","admin3");;
+  private ArrayList<String> adminList = new ArrayList<String>(Arrays.asList("admin","admin1","admin2","admin3"));
+
+  //private List<String> adminList = new ArrayList<String>["admin","admin1","admin2","admin3"];
   /**
    * Set up state for handling login-related requests. This method is only called when running in a
    * server, not when running in a test.
@@ -33,6 +37,11 @@ public class AdminServlet extends HttpServlet {
     setUserStore(UserStore.getInstance());
     setConversationStore(ConversationStore.getInstance());
     setMessageStore(MessageStore.getInstance());
+    //adminList.add("admin");
+    //adminList.add("admin1");
+    //adminList.add("admin2");
+    //adminList.add("admin3");
+
   }
 
   /**
@@ -105,7 +114,7 @@ public class AdminServlet extends HttpServlet {
       request.setAttribute("newestUser", newestUser);
       request.setAttribute("mostActiveUser", mostActiveUser);
       request.getRequestDispatcher("/WEB-INF/view/admin.jsp").forward(request, response);
-        }
+    }
 
 
 }
--- a/src/main/java/codeu/controller/AdminServlet.java
+++ b/src/main/java/codeu/controller/AdminServlet.java
@@ -61,6 +61,7 @@ public class AdminServlet extends HttpServlet {
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response)
       throws IOException, ServletException {
+
       String username = (String)request.getSession().getAttribute("user");
       if (username == null) {
         // user is not logged in, don't let them access the admin page
@@ -73,8 +74,8 @@ public class AdminServlet extends HttpServlet {
         // user was not found, don't let them access the admin page
         response.sendRedirect("/");
         return;
-      }      
-      //adminList.add("admin");
+      }
+      adminList.add("admin");
       if (user.getIsAdmin() == true) {
         if (!adminList.contains(user.getName())){
           adminList.add(user.getName());
--- a/src/main/java/codeu/controller/AdminServlet.java
+++ b/src/main/java/codeu/controller/AdminServlet.java
@@ -21,7 +21,7 @@ public class AdminServlet extends HttpServlet {
   private UserStore userStore;
   private ConversationStore conversationStore;
   private MessageStore messageStore;
-  private List<String> adminList = new ArrayList<>();
+  private List<String> adminList = List.of("admin","admin1", "admin2","admin3");;
   /**
    * Set up state for handling login-related requests. This method is only called when running in a
    * server, not when running in a test.
@@ -54,10 +54,6 @@ public class AdminServlet extends HttpServlet {
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response)
       throws IOException, ServletException {
-        adminList.add("admin");
-        adminList.add("admin1");
-        adminList.add("admin2");
-        adminList.add("admin3");
 
       String username = (String)request.getSession().getAttribute("user");
       if (username == null) {
@@ -82,36 +78,33 @@ public class AdminServlet extends HttpServlet {
 
       /* an attempt to grab information from the stores to display on the page
       if the user is admin*/
-          List<User> userList = userStore.getAllUsers();
-          int numOfUsers = userStore.getNumOfUsers();
-          int numOfConvos = conversationStore.getNumOfConversations();
-          int numOfMessages = messageStore.getNumOfMessages();
+      List<User> userList = userStore.getAllUsers();
+      int numOfUsers = userStore.getNumOfUsers();
+      int numOfConvos = conversationStore.getNumOfConversations();
+      int numOfMessages = messageStore.getNumOfMessages();
 
-          request.setAttribute("numOfUsers", numOfUsers);
-          request.setAttribute("numOfMessages", numOfMessages);
-          request.setAttribute("numOfConvos", numOfConvos);
+      request.setAttribute("numOfUsers", numOfUsers);
+      request.setAttribute("numOfMessages", numOfMessages);
+      request.setAttribute("numOfConvos", numOfConvos);
 
-          int mostMessages=0;
-          String newestUser = userList.get(userList.size()-1).getName();
-          String mostActiveUser = "";
-          int numOfAdmins = adminList.size();
+      int mostMessages=0;
+      String newestUser = userList.get(userList.size()-1).getName();
+      String mostActiveUser = "";
+      int numOfAdmins = adminList.size();
 
-          for(int i =0; i< numOfUsers; i++){
-            UUID userId = userList.get(i).getId();
-            int userMessagesListSize = messageStore.getMessagesFromAuthorSize(userId);
-            if (userMessagesListSize > mostMessages){
-              mostMessages = userMessagesListSize;
-              User activeUser = userStore.getUser(userId);
-              mostActiveUser = activeUser.getName();
-            } else if(mostActiveUser == null){
-              mostActiveUser = "We need activity";
-            }
+      for (User userIsActive : userStore.getAllUsers()) {
+        UUID userId = userIsActive.getId();
+        int userMessagesListSize = messageStore.getMessagesFromAuthor(userId).size();
+        if (userMessagesListSize > mostMessages){
+            mostMessages = userMessagesListSize;
+            mostActiveUser = userIsActive.getName();
           }
+      }
 
-            request.setAttribute("numOfAdmins", numOfAdmins);
-            request.setAttribute("newestUser", newestUser);
-            request.setAttribute("mostActiveUser", mostActiveUser);
-            request.getRequestDispatcher("/WEB-INF/view/admin.jsp").forward(request, response);
+      request.setAttribute("numOfAdmins", numOfAdmins);
+      request.setAttribute("newestUser", newestUser);
+      request.setAttribute("mostActiveUser", mostActiveUser);
+      request.getRequestDispatcher("/WEB-INF/view/admin.jsp").forward(request, response);
         }
 
 
--- a/src/main/java/codeu/controller/AdminServlet.java
+++ b/src/main/java/codeu/controller/AdminServlet.java
@@ -60,7 +60,7 @@ public class AdminServlet extends HttpServlet {
       }
 
       User user = UserStore.getUser(username);
-      
+
       if (user == null) {
         // user was not found, don't let them access the admin page
         System.out.println("User not found: " + username);
@@ -74,6 +74,7 @@ public class AdminServlet extends HttpServlet {
           int numOfConvos = ConversationStore.getInstance().getAllConversations().size();
           int numOfMessages = MessageStore.getInstance().getAllMessages().size();
           List<User> userList = UserStore.getInstance().getAllUsers();
+          //TODO additonal stats
           //String mostActiveUser = userList.get
           //String newestUser = userList.
           //String wordiestUser = userList
--- a/src/main/java/codeu/controller/AdminServlet.java
+++ b/src/main/java/codeu/controller/AdminServlet.java
@@ -72,7 +72,7 @@ public class AdminServlet extends HttpServlet {
       /* an attempt to grab information from the stores to display on the page */
           int numOfUsers = UserStore.getInstance().getAllUsers().size();
           int numOfConvos = ConversationStore.getInstance().getAllConversations().size();
-          int numOfMessages = MessageStore.getInstance().getMessages().size();
+          int numOfMessages = MessageStore.getInstance().getAllMessages().size();
           List<User> userList = UserStore.getInstance().getAllUsers();
           //String mostActiveUser = userList.get
           //String newestUser = userList.
--- a/src/main/java/codeu/controller/AdminServlet.java
+++ b/src/main/java/codeu/controller/AdminServlet.java
@@ -92,28 +92,24 @@ public class AdminServlet extends HttpServlet {
           request.setAttribute("numOfConvos", numOfConvos);
 
           int mostMessages=0;
-          User newestUser = userList.get(userStore.getNumOfUsers()-1);//.getName();
+          String newestUser = userList.get(userList.size()-1).getName();
           String mostActiveUser = "";
           int numOfAdmins = adminList.size();
 
           for(int i =0; i< numOfUsers; i++){
             UUID userId = userList.get(i).getId();
-            int userMessagesListSize = messageStore.getMessagesFromAuthor(userId).size();
-            if( userList.get(i) <= newestUser.getCreationTime()){
-              int userTime = userList.get(i).getCreationTime();
-            }
+            int userMessagesListSize = messageStore.getMessagesFromAuthorSize(userId);
             if (userMessagesListSize > mostMessages){
               mostMessages = userMessagesListSize;
-              mostActiveUser = userStore.getUser(userId).getName();
-            }else if(mostActiveUser == null){
+              User activeUser = userStore.getUser(userId);
+              mostActiveUser = activeUser.getName();
+            } else if(mostActiveUser == null){
               mostActiveUser = "We need activity";
             }
-            }
-
-            request.setAttribute("timeCreated",userTime )
+          }
 
             request.setAttribute("numOfAdmins", numOfAdmins);
-            //request.setAttribute("newestUser", newestUser);
+            request.setAttribute("newestUser", newestUser);
             request.setAttribute("mostActiveUser", mostActiveUser);
             request.getRequestDispatcher("/WEB-INF/view/admin.jsp").forward(request, response);
         }
--- a/src/main/java/codeu/controller/AdminServlet.java
+++ b/src/main/java/codeu/controller/AdminServlet.java
@@ -31,7 +31,7 @@ public class AdminServlet extends HttpServlet {
     super.init();
     setUserStore(UserStore.getInstance());
     setConversationStore(ConversationStore.getInstance());
-    setMessageStore(MessageStore.getInstance());xs
+    setMessageStore(MessageStore.getInstance());
   }
 
   /**
--- a/src/main/java/codeu/controller/AdminServlet.java
+++ b/src/main/java/codeu/controller/AdminServlet.java
@@ -23,7 +23,7 @@ public class AdminServlet extends HttpServlet {
   private UserStore userStore;
   private ConversationStore conversationStore;
   private MessageStore messageStore;
-  private ArrayList<String> adminList = new ArrayList<String>(Arrays.asList("admin","admin1","admin2","admin3"));
+  private List<String> adminList = Arrays.asList("admin","admin1","admin2","admin3");
 
 
   @Override
--- a/src/main/java/codeu/controller/AdminServlet.java
+++ b/src/main/java/codeu/controller/AdminServlet.java
@@ -73,9 +73,9 @@ public class AdminServlet extends HttpServlet {
         return;
       }
       User user = userStore.getUser(username);
-      if (adminList.contains(username){
+      if (adminList.contains(username)){
         User adminUser = userStore.getUser(username);
-        adminUser.setisAdmin(true);
+        adminUser.setIsAdmin(true);
       }
 
       if (user == null) {
@@ -83,10 +83,12 @@ public class AdminServlet extends HttpServlet {
         response.sendRedirect("/");
         return;
       }
+      if (user.getIsAdmin() == false) {
+        /* rediects user to homepage if not admin*/
+        response.sendRedirect("/");
+        return;
+      } else if(user.getIsAdmin()==true){
 
-
-      request.setAttribute("adminList", adminList);
-      if (adminList.contains(user.getName())) {
       /* an attempt to grab information from the stores to display on the page
       if the user is admin*/
           List<User> userList = userStore.getAllUsers();
@@ -106,26 +108,14 @@ public class AdminServlet extends HttpServlet {
             if (messageStore.getMessagesFromAuthor(userId).size() > mostMessages){
               mostMessages = messageStore.getMessagesFromAuthor(userId).size();
               mostActiveUser = userStore.getInstance().getUser(userId).getName();
-            }
-            if(userInfo.getIsAdmin() == true){
-              numOfAdmins ++;
-            }
-            if(mostActiveUser == null){
+            }else if(mostActiveUser == null){
               mostActiveUser = "We need activity";
             }
             }
-          }
+            request.setAttribute("newestUser", newestUser);
+            request.setAttribute("mostActiveUser", mostActiveUser);
+            request.getRequestDispatcher("/WEB-INF/view/admin.jsp").forward(request, response);
+        }
+    }
 
-
-          }
-          request.setAttribute("newestUser", newestUser);
-          request.setAttribute("mostActiveUser", mostActiveUser);
-          request.setAttribute("numOfAdmins", numOfAdmins);
-          request.getRequestDispatcher("/WEB-INF/view/admin.jsp").forward(request, response);
-
-        } else if (!adminList.contains(user.getName())) {
-            /* rediects user to homepage if not admin*/
-          response.sendRedirect("/");
-          return;
-  }
-}}
+}
--- a/src/main/java/codeu/controller/AdminServlet.java
+++ b/src/main/java/codeu/controller/AdminServlet.java
@@ -25,23 +25,13 @@ public class AdminServlet extends HttpServlet {
   private MessageStore messageStore;
   private ArrayList<String> adminList = new ArrayList<String>(Arrays.asList("admin","admin1","admin2","admin3"));
 
-  //private List<String> adminList = new ArrayList<String>["admin","admin1","admin2","admin3"];
-  /**
-   * Set up state for handling login-related requests. This method is only called when running in a
-   * server, not when running in a test.
-   */
 
   @Override
   public void init() throws ServletException {
     super.init();
     setUserStore(UserStore.getInstance());
     setConversationStore(ConversationStore.getInstance());
-    setMessageStore(MessageStore.getInstance());
-    //adminList.add("admin");
-    //adminList.add("admin1");
-    //adminList.add("admin2");
-    //adminList.add("admin3");
-
+    setMessageStore(MessageStore.getInstance());xs
   }
 
   /**
--- a/src/main/java/codeu/controller/AdminServlet.java
+++ b/src/main/java/codeu/controller/AdminServlet.java
@@ -19,8 +19,8 @@ import java.util.UUID;
 public class AdminServlet extends HttpServlet {
 
   private UserStore userStore;
-  private ConversationStore ConversationStore;
-  private MessageStore MessageStore;
+  private ConversationStore conversationStore;
+  private MessageStore messageStore;
   private List<String> adminList = new ArrayList<String>();
   /**
    * Set up state for handling login-related requests. This method is only called when running in a
@@ -30,7 +30,7 @@ public class AdminServlet extends HttpServlet {
   @Override
   public void init() throws ServletException {
     super.init();
-    setUserStore(userStore.getInstance());
+    setUserStore(UserStore.getInstance());
     setConversationStore(ConversationStore.getInstance());
     setMessageStore(MessageStore.getInstance());
   }
@@ -43,12 +43,12 @@ public class AdminServlet extends HttpServlet {
     this.userStore = userStore;
   }
 
-  void setConversationStore(ConversationStore ConversationStore) {
-    this.ConversationStore = ConversationStore;
+  void setConversationStore(ConversationStore conversationStore) {
+    this.conversationStore = conversationStore;
   }
 
-  void setMessageStore(MessageStore MessageStore) {
-    this.MessageStore = MessageStore;
+  void setMessageStore(MessageStore messageStore) {
+    this.messageStore = messageStore;
   }
 
   @Override
@@ -66,13 +66,9 @@ public class AdminServlet extends HttpServlet {
         return;
       }
       User user = userStore.getUser(username);
-<<<<<<< HEAD
       if (adminList.contains(username)){
         user.setIsAdmin(true);
       }
-
-=======
->>>>>>> parent of afb1bd2... added admin tab to navbar. Admin tab should only show when user is an admin
       if (user == null) {
         // user was not found, don't let them access the admin page
         response.sendRedirect("/");
--- a/src/main/java/codeu/controller/AdminServlet.java
+++ b/src/main/java/codeu/controller/AdminServlet.java
@@ -74,8 +74,7 @@ public class AdminServlet extends HttpServlet {
       }
       User user = userStore.getUser(username);
       if (adminList.contains(username)){
-        User adminUser = userStore.getUser(username);
-        adminUser.setIsAdmin(true);
+        user.setIsAdmin(true);
       }
 
       if (user == null) {
@@ -87,7 +86,7 @@ public class AdminServlet extends HttpServlet {
         /* rediects user to homepage if not admin*/
         response.sendRedirect("/");
         return;
-      } else if(user.getIsAdmin()==true){
+      }
 
       /* an attempt to grab information from the stores to display on the page
       if the user is admin*/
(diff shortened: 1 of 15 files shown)


src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,93 @@
...
+  void setMessageStore(MessageStore MessageStore) {
+    this.MessageStore = MessageStore;
+  }
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response)
+      throws IOException, ServletException {
+      String username = (String)request.getSession().getAttribute("user");
+      if (username == null) {
+        // user is not logged in, don't let them create a conversation
```
vocatwk: is creating a conversation what they would've been able to do if it weren't null?

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,93 @@
...
+        response.sendRedirect("/");
+        return;
+      }
+
+      if (user.getAdmin() == true) {
+      /* an attempt to grab information from the stores to display on the page if the user is admin*/
+          int numOfUsers = UserStore.getInstance().getAllUsers().size();
+          int numOfConvos = ConversationStore.getInstance().getAllConversations().size();
+          int numOfMessages = MessageStore.getInstance().getAllMessages().size();
+          List<User> userList = UserStore.getInstance().getAllUsers();
```
vocatwk: You can do this in the beginning of if block, then call size on userList to get numOfUsers so you don't have to call those methods twice.

src/main/java/codeu/model/data/User.java
```diff
@@ -70,4 +71,11 @@ public void setAboutMe(String message) {
   public String getAboutMe() {
     return aboutMe;
   }
+
+  public void setAdmin(boolean a){
+    isAdmin = a;
+  }
+    public boolean getAdmin(){
```
vocatwk: indentation

src/main/webapp/WEB-INF/view/admin.jsp
```diff
@@ -0,0 +1,33 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>Admin Login</title>
+  <link rel="stylesheet" href="/css/main.css">
+</head>
+<body>
+
+  <nav>
```
vocatwk: You can use the navbar jsp when Guillermo merges those changes

src/test/java/codeu/controller/AdminServletTest.java
```diff
@@ -0,0 +1,115 @@
...
+        new Conversation(UUID.randomUUID(), UUID.randomUUID(), "test_conversation", Instant.now()));
+    Mockito.when(mockConversationStore.getAllConversations()).thenReturn(fakeConversationList);
+
+    List<Message> fakeMessageList = new ArrayList<>();
+    fakeMessageList.add(
+        new Message(UUID.randomUUID(),fakeConversationId, UUID.randomUUID(),"test_message",Instant.now()));
+    Mockito.when(mockMessageStore.getAllMessages()).thenReturn(fakeMessageList);
+    AdminServlet.doGet(mockRequest, mockResponse);
+
+    //Mockito.verify(mockRequestDispatcher).forward(mockRequest, mockResponse);
```
vocatwk: Maybe verify if it set those attributes and goes to admin.jsp.

vocatwk: Also, I'm not sure but in your doGet() doesn't it call getAttribute("user") on request object. How come you don't call when() for that and return the fake user you made to maybe check if it correctly identifies admins?

src/test/java/codeu/controller/AdminServletTest.java
```diff
@@ -0,0 +1,115 @@
...
+    Mockito.when(mockSession.getAttribute("user")).thenReturn("test_username");
+    Mockito.when(mockUserStore.getUser("test_username")).thenReturn(null);
+
+    AdminServlet.doGet(mockRequest, mockResponse);
+    Mockito.verify(mockResponse).sendRedirect("/");
+  }
+
+  @Test
+  public void testDoGet_isAdmin() throws IOException, ServletException {
+  Mockito.when(mockRequest.getSession().getAttribute("user")).thenReturn("test_username");
```
vocatwk: indentation

src/test/java/codeu/controller/AdminServletTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+    int fakenumOfUsers = mockUserStore.getAllUsers().size();
+    int fakenumOfConvos = mockConversationStore.getAllConversations().size();
+    int fakenumOfMessages = mockMessageStore.getAllMessages().size();
+
+    AdminServlet.doGet(mockRequest, mockResponse);
```
vocatwk: wait should you be verifying anything here? I also don't understand what this is really testing.

src/test/java/codeu/controller/AdminServletTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+    UUID fakeConversationId = UUID.randomUUID();
+    List<Conversation> fakeConversationList = new ArrayList<>();
+    fakeConversationList.add(
+        new Conversation(UUID.randomUUID(), UUID.randomUUID(), "test_conversation", Instant.now()));
+    Mockito.when(mockConversationStore.getAllConversations()).thenReturn(fakeConversationList);
+
+    List<Message> fakeMessageList = new ArrayList<>();
+    fakeMessageList.add(
+        new Message(UUID.randomUUID(),fakeConversationId, UUID.randomUUID(),"test_message",Instant.now()));
```
vocatwk: what is the point of fakeConversationId variable, it doesn't seem like you use it elsewhere?

vocatwk:
Your idea on how to add admins sounds good but I don't understand what you mean by not being able to access the page locally without being an admin. and for the admin tab, you can set the isAdmin attribute in session when admins log in and just use an if statement in the navbar.

yebrahim:
Nice!

Let's discuss how we're going to identify admins here. My thinking was that we should start out simple by maintaining a hard coded list:
```
List<String> adminsList = new ArrayList<>();
list.add("admin1");
list.add("admin2");
...
```
Then just checking if the username exists in this list.

Another option would be to insert some admin users in the database before the app starts.

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  /**
+   * Sets the UserStore used by this servlet. This function provides a common setup method for use
+   * by the test framework or the servlet's init() function.
+   */
+  void setUserStore(UserStore UserStore) {
+    this.UserStore = UserStore;
+  }
+
```
lealgamb: method doc.

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,93 @@
...
+  void setConversationStore(ConversationStore ConversationStore) {
+    this.ConversationStore = ConversationStore;
+  }
+
```
lealgamb: method doc.

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import org.mindrot.jbcrypt.BCrypt;
+
+/** Servlet class responsible for the Admin page. */
+public class AdminServlet extends HttpServlet {
+
+  private UserStore UserStore;
+  private ConversationStore ConversationStore;
+  private MessageStore MessageStore;
```
lealgamb: Consider camelCase for consistency w/ other files.

src/main/java/codeu/controller/AdminServlet.java
lealgamb: Same question. How is the admin page related to conversations?

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,93 @@
...
+        // user is not logged in, don't let them create a conversation
+        response.sendRedirect("/");
+        return;
+      }
+
+      User user = UserStore.getUser(username);
+
+      if (user == null) {
+        // user was not found, don't let them access the admin page
+        System.out.println("User not found: " + username);
```
lealgamb: Remember to delete for final product.

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,93 @@
...
+          //String wordiestUser = userList
+
+          request.setAttribute("numOfUsers", numOfUsers);
+          request.setAttribute("numOfMessages", numOfMessages);
+          request.setAttribute("numOfConvos", numOfConvos);
+          request.getRequestDispatcher("/WEB-INF/view/admin.jsp").forward(request, response);
+
+        } else if (user.getAdmin() == false) {
+            /* rediects user to homepage if not admin*/
+          System.out.println("User not an Admin: " + username);
```
lealgamb: same as before

src/main/java/codeu/model/data/User.java
```diff
@@ -70,4 +71,11 @@ public void setAboutMe(String message) {
   public String getAboutMe() {
     return aboutMe;
   }
+
+  public void setAdmin(boolean a){
```
lealgamb: nitpick: descriptive parameter name

jasminechau2:
I meant I hadn't added a way for people to become admins without hardcoding it at first. I will make changes and begin working on that.

yebrahim:
You'll want to resolve these conflicts in a separate commit before continuing to work on the changes.

src/main/webapp/WEB-INF/view/navbar.jsp
```diff
@@ -1,5 +1,6 @@
 <nav>
   <% String navBarUser = (String)request.getSession().getAttribute("user"); %>
+  <% List<String> adminList = (List<String>) request.getSession().getAttribute("adminList"); %>
```
lealgamb: Why do we want the list here?

src/test/java/codeu/controller/AdminServletTest.java
```diff
@@ -99,17 +93,42 @@ public void testDoGet_InvalidUser() throws IOException, ServletException {
...
-    int fakenumOfUsers = mockUserStore.getAllUsers().size();
-    int fakenumOfConvos = mockConversationStore.getAllConversations().size();
-    int fakenumOfMessages = mockMessageStore.getAllMessages().size();
+  public void testDoGet_AdminStatsPage() throws IOException, ServletException {
+    List<User> fakeUserList = new ArrayList<>();
+    User fakeUser = new User(UUID.randomUUID(), "test_username", "fakePasswordHash", Instant.now());
+    fakeUser.setAdmin(true);
+    fakeUserList.add(fakeUser);
+    Mockito.when(mockUserStore.getUser("test_username")).thenReturn(fakeUser);
+    List<String> fakeAdminList = new ArrayList<String>();
```
lealgamb: Consistency: Lines 97 vs. 102 (and in other areas)
`ArrayList<>` vs. `ArrayList<String>`

src/main/webapp/WEB-INF/view/navbar.jsp
jasminechau2: I was attempting to add the admin page as a tab only when the user is an admin, for some reason when I called adminList.contains(navBarUser) it would give me an error. I forgot to delete it from my attempts.

lealgamb: I see. We can try adding `boolean isAdmin` in the `User` entity then adding a condition to check for this.

jasminechau2: I do have an isAdmin boolean currently in the User. How would I go about using it in the navbar.jsp?

lealgamb: Sorry I missed that. You can use `UserStore` to get the `User` object with `navBarUser` as the parameter, then call on `isAdmin()`.

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,116 @@
...
+
+      String username = (String)request.getSession().getAttribute("user");
+      if (username == null) {
+        // user is not logged in, don't let them access the admin page
+        response.sendRedirect("/");
+        return;
+      }
+      User user = userStore.getUser(username);
+
+      if (user == null) {
```
lealgamb: Should we redirect them to the homepage or display some "Access Denied" page?

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,116 @@
...
+        // user was not found, don't let them access the admin page
+        response.sendRedirect("/");
+        return;
+      }
+      adminList.add("admin");
```
lealgamb: Why is this needed?

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,116 @@
...
+      if the user is admin*/
+          List<User> userList = UserStore.getInstance().getAllUsers();
+          int numOfUsers = userList.size();
+          int numOfConvos = conversationStore.getInstance().getAllConversations().size();
+          int numOfMessages = messageStore.getInstance().getAllMessages().size();
+          String newestUser = userList.get(userList.size()-1).getName();
+          String mostActiveUser = "";
+          for(int i =0; i< numOfUsers; i++){
+            UUID userId = userList.get(i).getId();
+              int mostMessages=0;
```
lealgamb: indentation

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,116 @@
...
+            if (messageStore.getMessagesFromAuthor(userId).size() > mostMessages){
```
lealgamb: Can we store the current size as a variable to make it cleaner?

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,116 @@
...
+              mostMessages = messageStore.getMessagesFromAuthor(userId).size();
+              mostActiveUser = userStore.getInstance().getUser(userId).getName();
```
lealgamb: How will private messages affect this? Say that User A has 50 messages, all private, and User B has 20 messages, all public. Do we want people to know User A is the most active?

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,116 @@
...
+      request.setAttribute("adminList", adminList);
+      if (adminList.contains(user.getName())) {
+      /* an attempt to grab information from the stores to display on the page
+      if the user is admin*/
+          List<User> userList = UserStore.getInstance().getAllUsers();
+          int numOfUsers = userList.size();
+          int numOfConvos = conversationStore.getInstance().getAllConversations().size();
+          int numOfMessages = messageStore.getInstance().getAllMessages().size();
+          String newestUser = userList.get(userList.size()-1).getName();
+          String mostActiveUser = "";
```
lealgamb: When the app is launched, `mostActiveUser` will be `""`. If that's the case, maybe we can display a message encouraging users to be active.

src/main/webapp/WEB-INF/view/admin.jsp
```diff
@@ -0,0 +1,23 @@
...
+  <%-- getting statistics  --%>
+	<br/>
+	<br/>
+  <div id="container">
+    <h1>Admin Page</h1>
+		<p> Here are some statistics of Nemo. </p>
+		<p> Current number of Users: <%= request.getAttribute("numOfUsers") %></p>
+		<p> Current number of Messages: <%= request.getAttribute("numOfMessages") %></p>
+		<p> Current number of Conversations: <%= request.getAttribute("numOfConvos") %></p>
+    <p> Most active user: <%= request.getAttribute("mostActiveUser") %></p>
```
lealgamb: Keep indentation consistent.

src/main/java/codeu/controller/AdminServlet.java
yebrahim: This will probably need a separate change if we think private messages shouldn't count. I'd open an issue to track it separately.

src/main/java/codeu/controller/AdminServlet.java
jasminechau2: I can do that

src/main/java/codeu/controller/AdminServlet.java
jasminechau2: Once the private messages are implemented I can change the stats to show the amount of public and private messages as well total messages. I can from then make most active user based on public messages if we store the two types in different list.

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -12,14 +12,16 @@
...
 public class AdminServlet extends HttpServlet {
 
-  private UserStore UserStore;
-  private ConversationStore ConversationStore;
-  private MessageStore MessageStore;
-
+  private UserStore userStore;
+  private ConversationStore conversationStore;
+  private MessageStore messageStore;
+  private List<String> adminList = new ArrayList<>();
```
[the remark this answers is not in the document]
vocatwk: I'm also a bit confused on why you're maintaining a list of admins since you're checking if the user is an admin in the beginning and adding him to the admin list. Why not always just check if the user is an admin by using getIsAdmin throughout. Wouldn't it be less expensive than searching through a list?

src/main/java/codeu/model/data/User.java
```diff
@@ -72,10 +72,13 @@ public String getAboutMe() {
...
+  /** Sets if the user is admin. */
+  public void setIsAdmin(boolean a){
     isAdmin = a;
   }
-    public boolean getAdmin(){
+
+  /** Returns true if the user is admin. Returns false if not */
+  public boolean getIsAdmin(){
       return isAdmin;
 }
```
vocatwk: indentation

src/test/java/codeu/controller/AdminServletTest.java
```diff
@@ -59,28 +59,63 @@ public void setup() {
...
+
     AdminServlet.doGet(mockRequest, mockResponse);
 
-    //Mockito.verify(mockRequestDispatcher).forward(mockRequest, mockResponse);
+    List<String> fakeAdminList = new ArrayList<>();
 
+    Mockito.when(mockSession.getAttribute("user")).thenReturn("test_username");
+    Mockito.when(mockUserStore.getUser("test_username")).thenReturn(fakeUser);
+
+      if(fakeUser.getIsAdmin() == true ){
```
vocatwk: indentation

vocatwk: also fakeUser.getIsAdmin() is always going to return false right. What is the point of this if block?

src/test/java/codeu/controller/AdminServletTest.java
```diff
@@ -59,28 +59,63 @@ public void setup() {
...
+        if(fakeAdminList.contains(fakeUser.getName())){
+        fakeAdminList.add(fakeUser.getName());
+        }
+      }
+    if(fakeAdminList.contains(fakeUser.getName())){
```
vocatwk: Same as above with this if block.

yebrahim:
I can see the problem of figuring out if a user is admin is complicating things quite a bit. Can we start with a simple implementation that hard codes a list of usernames as admins? Something like:
`List<string> adminUsernames = List.of("admin1", "admin2");`

Then in your servlet, you can check if the current session's username is in this list, and redirect to "/" if it isn't.

We can later build a more sophisticated enrollment mechanism for admins if we want. :)

jasminechau2:
The code handles that by checking if the getIsAdmin is true or false. If it is true they have access to the page and if it is false it redirects them to the front page without ever letting them see the page. I also implemented it so that the navbar only shows the admin page if the user is already and admin

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,169 @@
...
+import java.time.Instant;
+
+
+/** Servlet class responsible for the Admin page. */
+public class AdminServlet extends HttpServlet {
+
+  private UserStore userStore;
+  private ConversationStore conversationStore;
+  private MessageStore messageStore;
+  private List<String> nameOfUserWhoWantToBeAdmin = new ArrayList<>();
```
yebrahim: Can we revert all of the changes that introduced this?

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,169 @@
...
+        User userAdmin = userStore.getUser(username);
+        user.setIsAdmin(true);
+      }
+
+      if(user == null) {
+        // user was not found, don't let them access the admin page
+        response.sendRedirect("/");
+        return;
+      }
+      //user.setIsAdmin(true);
```
lealgamb: contradicting comment

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,169 @@
...
+      if(user.getIsAdmin() == false) {
+        response.sendRedirect("/");
+        return;
+      }
+      /* an attempt to grab information from the stores to display on the page
+      if the user is admin*/
+
+        int numOfUsers = UserStore.getInstance().amountOfUsers();
```
lealgamb: You already have instances of all the store objects. Also, for the UserStore you're using the class rather than the object you created.

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,141 @@
...
+      throws IOException, ServletException {
+      String username = (String)request.getSession().getAttribute("user");
+
+      if(username == null) {
+        // user is not logged in, don't let them access the admin page
+        response.sendRedirect("/");
+        return;
+      }
+      List<String> adminUsernames = new ArrayList<String>();
+      adminUsernames.add("admin");
```
lealgamb: Sorry, I still don't see why these are added.

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,141 @@
...
+        int numOfMessages = messageStore.amountOfMessages();
+
+        request.setAttribute("numOfUsers", numOfUsers);
+        request.setAttribute("numOfConvos", numOfConvos);
+        request.setAttribute("numOfMessages", numOfMessages);
+
+        List<User> userList = userStore.getAllUsers();
+        String newestUser ="";
+        String mostActiveUser = null;
+        int numOfAdmins = 0;
```
lealgamb: Why don't we set this to the size of the list?

src/main/webapp/WEB-INF/view/admin.jsp
```diff
@@ -0,0 +1,26 @@
...
+  <link rel="stylesheet" href="/css/main.css">
+</head>
+<body>
+  <%@ include file="navbar.jsp" %>
+  <%-- getting statistics  --%>
+	<br/>
+	<br/>
+  <div id="container">
+    <h1>Admin Page</h1>
+		<p> Here are some statistics of Nemo. </p>
```
lealgamb: Keep indentation consistent

src/main/java/codeu/controller/AdminServlet.java
jasminechau2: It's a hardcoded list of admins so that some users already have access to the page.

lealgamb:
Something to keep in mind for later is to update the store by adding the isAdmin variable.

src/main/java/codeu/controller/AdminServlet.java
lealgamb: Ahh I see. How will we make sure the User objects are in the datastore once the app launches? I'm a little lost on that part.

jasminechau2:
The UserStore?

lealgamb:
I was thinking of `PersistentDataStore`'s `loadUsers()` and `writeThrough(User user)`, then updating the tests if needed.

jasminechau2:
okay I can do that

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,142 @@
...
+        response.sendRedirect("/");
+        return;
+      }
+      if(user.getIsAdmin() == false) {
+        response.sendRedirect("/");
+        return;
+      }
+      /* an attempt to grab information from the stores to display on the page
+      if the user is admin*/
+      if(!adminUsernames.contains(username)){
```
vocatwk: if adminUsernames doesn't contain username then the user would've been redirected to the index page right, so does this if block ever execute?

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,142 @@
...
+            newestUser = userList.get(numOfUsers-1).getName();
+        }
+
+        for(int i =0; i< numOfUsers; i++){
+          User userInfo = userList.get(i);
+          UUID userId = userInfo.getId();
+          int userMessagesListSize = messageStore.getMessagesFromAuthor(userId).size();
+          int mostMessages=0;
+          if (userMessagesListSize > mostMessages){
+            mostMessages = messageStore.getMessagesFromAuthor(userId).size();
```
vocatwk: you can just set mostMessages = userMessagesListSize, instead of calling the methods again.

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,142 @@
...
+        if (userList.size()>0){
+            newestUser = userList.get(numOfUsers-1).getName();
+        }
+
+        for(int i =0; i< numOfUsers; i++){
+          User userInfo = userList.get(i);
+          UUID userId = userInfo.getId();
+          int userMessagesListSize = messageStore.getMessagesFromAuthor(userId).size();
+          int mostMessages=0;
+          if (userMessagesListSize > mostMessages){
```
vocatwk: if you set mostMessage = 0 inside the loop then this if block is going to be true for any user that has sent atleast one message. So the most active user will just be the last person to send a messge, right?

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,142 @@
...
+            mostMessages = messageStore.getMessagesFromAuthor(userId).size();
+            mostActiveUser = userStore.getInstance().getUser(userId).getName();
+          }
+          if(userInfo.getIsAdmin() == true){
+            numOfAdmins ++;
```
vocatwk: I'm confused. I thought the number of admins was fixed?

src/test/java/codeu/controller/AdminServletTest.java
```diff
@@ -0,0 +1,184 @@
...
+      List<Conversation> fakeConversationList = new ArrayList<>();
+      fakeConversationList.add(
+          new Conversation(UUID.randomUUID(), UUID.randomUUID(), "test_conversation", Instant.now()));
+        Mockito.when(mockConversationStore.getAllConversations()).thenReturn(fakeConversationList);
+
+      List<Message> fakeMessageList = new ArrayList<>();
+      fakeMessageList.add(
+          new Message(UUID.randomUUID(),UUID.randomUUID(), UUID.randomUUID(),"test_message",Instant.now()));
+        Mockito.when(mockMessageStore.getAllMessages()).thenReturn(fakeMessageList);
+
```
vocatwk: Shouldn't you be calling the doGet method? Also, maybe verify that it sets the attributes like number of users and such or if it redirects according to what should happen in this situation.

src/test/java/codeu/controller/AdminServletTest.java
```diff
@@ -0,0 +1,184 @@
...
+      mockUser = Mockito.mock(User.class);
+      mockConversation = Mockito.mock(Conversation.class);
+    }
+
+  @Test
+  public void testDoGet() throws IOException, ServletException {
+
+      List<User> fakeUserList = new ArrayList<>();
+      User fakeUser = new User( UUID.randomUUID(), "test_username", "fakePasswordHash", Instant.now());
+      UUID fakeUserId = UUID.randomUUID();
```
vocatwk: Do you use this?

src/test/java/codeu/controller/AdminServletTest.java
```diff
@@ -0,0 +1,184 @@
...
+    Mockito.when(mockUserStore.getUser("test_username")).thenReturn(mockUser);
+    AdminServlet.doGet(mockRequest, mockResponse);
+    Mockito.when(mockUser.getIsAdmin()).thenReturn(false);
+    Mockito.verify(mockResponse).sendRedirect("/");
+  }
+
+  public void testDoGet_getStats() throws IOException, ServletException{
+    List<User> fakeUserList = new ArrayList<>();
+    User fakeUser = new User( UUID.randomUUID(), "test_username", "fakePasswordHash", Instant.now());
+    UUID fakeUserId = UUID.randomUUID();
```
vocatwk: Do you use this?

src/test/java/codeu/controller/AdminServletTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+    List<Message> fakeMessageList = new ArrayList<>();
+    fakeMessageList.add(
+        new Message(UUID.randomUUID(),UUID.randomUUID(), UUID.randomUUID(),"test_message",Instant.now()));
+      Mockito.when(mockMessageStore.getAllMessages()).thenReturn(fakeMessageList);
+
+
+    Mockito.when(mockSession.getAttribute("user")).thenReturn("test_username");
+    Mockito.when(mockUserStore.getUser("test_username")).thenReturn(mockUser);
+    Mockito.when(mockUserStore.getAllUsers()).thenReturn(fakeUserList);
```
vocatwk: already set on line 123

src/test/java/codeu/controller/AdminServletTest.java
```diff
@@ -0,0 +1,184 @@
...
+    Mockito.verify(mockUserStore).amountOfUsers();
+    Mockito.verify(mockConversationStore).amountOfConnversations();
+    Mockito.verify(mockMessageStore).amountOfMessages();
+
+    Mockito.verify(mockRequest).setAttribute("numOfUsers", fakeUserList.size());
+    Mockito.verify(mockRequest).setAttribute("numOfConvos", fakeMessageList.size());
+    Mockito.verify(mockRequest).setAttribute("numOfMessages", fakeConversationList.size());
+    Mockito.verify(mockRequestDispatcher).forward(mockRequest, mockResponse);
+
+    Mockito.when(mockSession.getAttribute("user")).thenReturn("test_username");
```
vocatwk: I think you've already done 155- 160. or is that on purpose?

src/test/java/codeu/controller/AdminServletTest.java
```diff
@@ -0,0 +1,184 @@
...
+    Mockito.when(mockUser.getIsAdmin()).thenReturn(true);
+    Mockito.when(mockUserStore.amountOfUsers()).thenReturn(fakeUserList.size());
+
+    AdminServlet.doGet(mockRequest, mockResponse);
+
+    String fakeNewestUser = fakeUserList.get(fakeUserList.size()-1).getName();
+    String fakeMostActiveUser = "";
+    int fakeNumOfAdmins = 0;
+
+    for(int i =0; i< fakeUserList.size(); i++){
```
vocatwk: I don't understand what you're trying to do with the code from 166- 175

src/test/java/codeu/controller/AdminServletTest.java
```diff
@@ -0,0 +1,184 @@
...
+  }
+  @Test
+  public void testDoGet_isAdminFalse() throws IOException, ServletException{
+    Mockito.when(mockSession.getAttribute("user")).thenReturn("test_username");
+    Mockito.when(mockUserStore.getUser("test_username")).thenReturn(mockUser);
+    AdminServlet.doGet(mockRequest, mockResponse);
+    Mockito.when(mockUser.getIsAdmin()).thenReturn(false);
+    Mockito.verify(mockResponse).sendRedirect("/");
+  }
+
```
vocatwk: Did you mean to put @Test here?

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,142 @@
...
+        return;
+      }
+      List<String> adminUsernames = new ArrayList<>();
+      adminUsernames.add("admin");
+      adminUsernames.add("admin1");
+      adminUsernames.add("admin2");
+      adminUsernames.add("admin3");
+
+
+      User user = userStore.getUser(username);
```
lealgamb: Why don't we just check if user is an admin with the method from the User class?

src/main/java/codeu/controller/AdminServlet.java
lealgamb: It's better to initialize it before the for loop.

src/main/java/codeu/controller/AdminServlet.java
lealgamb: Also, we isn't numOfAdmins just adminUsernames.size()?

src/main/webapp/WEB-INF/view/navbar.jsp
```diff
@@ -10,4 +14,7 @@
   <% } %>
   <a href="/about.jsp">About</a>
   <a href="/activityfeed">Activity Feed</a>
-</nav>
\ No newline at end of file
+  <%if(navBarUser != null && ifAdmin.getIsAdmin() == true){%>
```
lealgamb: It might be cleaner to just have one variable for navBarUser and ifAdmin, say `User user` and checking if it isn't null and if it's an admin.

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  void setMessageStore(MessageStore messageStore) {
+    this.messageStore = messageStore;
+  }
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response)
+      throws IOException, ServletException {
+        adminList.add("admin");
```
yebrahim: Is this the best place to initialize the list? I think this means that every time there's a GET request, these four admins will be added to the list. We should find a spot that gets called only once.

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,118 @@
...
+      }
+      if (user.getIsAdmin() == false) {
+        /* rediects user to homepage if not admin*/
+        response.sendRedirect("/");
+        return;
+      }
+
+      /* an attempt to grab information from the stores to display on the page
+      if the user is admin*/
+          List<User> userList = userStore.getAllUsers();
```
yebrahim: Indentation.

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,118 @@
...
+          request.setAttribute("numOfUsers", numOfUsers);
+          request.setAttribute("numOfMessages", numOfMessages);
+          request.setAttribute("numOfConvos", numOfConvos);
+
+          int mostMessages=0;
+          String newestUser = userList.get(userList.size()-1).getName();
+          String mostActiveUser = "";
+          int numOfAdmins = adminList.size();
+
+          for(int i =0; i< numOfUsers; i++){
```
yebrahim: nitpick: looks like you could also do:
```
for (User user : userStore.getAllUsers()) {
  UUID userId = user.getId();
```

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -0,0 +1,118 @@
...
+            UUID userId = userList.get(i).getId();
+            int userMessagesListSize = messageStore.getMessagesFromAuthorSize(userId);
+            if (userMessagesListSize > mostMessages){
+              mostMessages = userMessagesListSize;
+              User activeUser = userStore.getUser(userId);
+              mostActiveUser = activeUser.getName();
+            } else if(mostActiveUser == null){
```
yebrahim: I think you can get rid of this else if entirely. `mostActiveUser` shouldn't ever be `null`, and we can instead check if it's empty in the JSP and not display anything.

src/main/java/codeu/model/store/basic/MessageStore.java
```diff
@@ -109,4 +109,21 @@ public void addMessage(Message message) {
   public void setMessages(List<Message> messages) {
     this.messages = messages;
   }
+
+  public int getNumOfMessages(){
+    return messages.size();
+  }
+
+  public int getMessagesFromAuthorSize(UUID authorId) {
```
yebrahim: This method can just call `getMessagesFromAuthor`, then return its size.
Or you could just get rid of this method, and have the caller call `getMessagesFromAuthor('name').size()`?

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -21,7 +23,9 @@
   private UserStore userStore;
   private ConversationStore conversationStore;
   private MessageStore messageStore;
-  private List<String> adminList = List.of("admin","admin1", "admin2","admin3");;
+  private ArrayList<String> adminList = new ArrayList<String>(Arrays.asList("admin","admin1","admin2","admin3"));
```
yebrahim: Does the old line not work as well?

src/main/java/codeu/controller/AdminServlet.java
```diff
@@ -21,7 +23,9 @@
...
+
+  //private List<String> adminList = new ArrayList<String>["admin","admin1","admin2","admin3"];
```
yebrahim: Remove commented lines, same below.

src/main/java/codeu/controller/AdminServlet.java
yebrahim: Can you try: `private List<String> adminList = Arrays.asList(".....`?

yebrahim:
@jasminechau2 now that this is merged, can you add a little section in the README file about it? Just a few words saying we have this page and what it includes, nothing wordy.

jasminechau2:
ok

PR #36, opened 2018-05-31
Added tests for profile pages MVP

vocatwk:
I added 3 tests: where user is logged in and navigating to other users' profiles and their own profile, and one where user is not logged in.

### DIFF
--- a/src/test/java/codeu/controller/ProfileServletTest.java
+++ b/src/test/java/codeu/controller/ProfileServletTest.java
@@ -13,6 +13,7 @@ import javax.servlet.RequestDispatcher;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.Mockito;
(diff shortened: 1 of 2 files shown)


src/test/java/codeu/controller/ProfileServletTest.java
```diff
@@ -13,6 +13,7 @@
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
```
lealgamb: Sorry, I still don't completely understand what a session is so can you please explain how you're testing this.

lealgamb: Sorry, I still don't completely understand what a session is so can you please explain how you're testing this.

vocatwk: Since the doPost calls getSession() and then getAttribute() on the session it gets from the first method to get the name of the logged in user(if any), I just made a mockSession and used the when() method to simulate the calls.

lealgamb: I see, thanks.

PR #38, opened 2018-06-02
Implemented EventStore Class 

lealgamb:
I created an EventStore class. I then updated the Event class to be compatible with the updates in PersistentStorageAgent and PersistentDataStore. 

I was receiving a warning for casting an Object type as a List<String> but I suppressed it. I'm pretty sure I was casting correctly, but let me know what you think.

### DIFF
--- a/src/main/java/codeu/model/data/Event.java
+++ b/src/main/java/codeu/model/data/Event.java
@@ -1,10 +1,5 @@
 package codeu.model.data;
 
-import codeu.model.data.User;
-import codeu.model.data.Conversation;
-import codeu.model.data.Message;
-import codeu.model.store.basic.UserStore;
-import codeu.model.store.basic.ConversationStore;
 import java.time.Instant;
 import java.util.UUID;
 import java.util.List;
@@ -16,6 +11,7 @@ import java.util.ArrayList;
  */
 public class Event {
 
+  private final UUID id;
   private final String type;
   private final Instant creation;
   /** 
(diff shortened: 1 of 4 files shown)


src/main/java/codeu/model/data/Event.java
```diff
@@ -26,54 +22,27 @@
...
-      information.add(conversation.getTitle());
-    }
-    else if (type.equals("Message")) {
-      Message message = (Message)classType;
-      information.add(userStore.getUser(message.getAuthorId()).getName());
-      information.add(conversationStore.getConversation(message.getConversationId()).getTitle());
-      information.add(message.getContent());
-    }
+    this.information = new ArrayList<>();
+    for (String entry : information)
```
yebrahim: Isn't this equivalent to just setting `this.information = information`?

yebrahim:
The cast looks fine to me, although I'm not a Java expert. :)

Good work on adding this abstraction layer, it makes things a lot clearer around adding and sorting events.

src/main/java/codeu/model/data/Event.java
lealgamb: I did it this way thinking of shallow vs. deep copy, but I may have been wrong. Should I change it?

yebrahim: But do we need a deep copy? Do you plan for the caller of the constructor to modify the List object after passing it in?

PR #45, opened 2018-06-07
Added an option to let users make conversations private

### LINKED ISSUE #42: Add option to mark conversation as private

vocatwk:
I have added a settings button next to the conversation title in the chat page(I haven't paid attention to css yet) that shows an option to make the conversation private when you click it. When users click this button, the conversation, if not already private, becomes private. Nothing happens otherwise.

screen shot 2018-06-07 at 1 14 02 am

I have also added 2 test for the extra code in the doPost method of ChatServlet, to test the conditions in which the user is trying to make a public or an already private conversation method, private.

resolves #42

### DIFF
--- a/src/main/java/codeu/controller/ChatServlet.java
+++ b/src/main/java/codeu/controller/ChatServlet.java
@@ -182,14 +182,14 @@ public class ChatServlet extends HttpServlet {
               Instant.now());
 
       messageStore.addMessage(message);
-    }
 
-    List<String> messageInformation = new ArrayList<>();
-    messageInformation.add(user.getName());
-    messageInformation.add(conversationTitle);
-    messageInformation.add(cleanedMessageContent);
-    Event messageEvent = new Event(UUID.randomUUID(), "Message", message.getCreationTime(), messageInformation);
-    eventStore.addEvent(messageEvent);
+      List<String> messageInformation = new ArrayList<>();
+      messageInformation.add(user.getName());
+      messageInformation.add(conversationTitle);
+      messageInformation.add(cleanedMessageContent);
+      Event messageEvent = new Event(UUID.randomUUID(), "Message", message.getCreationTime(), messageInformation);
+      eventStore.addEvent(messageEvent);
+    }
 
     // redirect to a GET request
     response.sendRedirect("/chat/" + conversationTitle);
--- a/src/main/java/codeu/controller/ChatServlet.java
+++ b/src/main/java/codeu/controller/ChatServlet.java
@@ -117,6 +117,7 @@ public class ChatServlet extends HttpServlet {
     String username = (String) request.getSession().getAttribute("user");
     if (username == null) {
       // user is not logged in, don't let them add a message
+      // or make conversation private
       response.sendRedirect("/login");
       return;
     }
@@ -124,6 +125,7 @@ public class ChatServlet extends HttpServlet {
     User user = userStore.getUser(username);
     if (user == null) {
       // user was not found, don't let them add a message
+      // or make conversation private
       response.sendRedirect("/login");
       return;
     }
@@ -138,20 +140,31 @@ public class ChatServlet extends HttpServlet {
       return;
     }
 
+    String type = request.getParameter("type");
     String messageContent = request.getParameter("message");
 
-    // this removes any HTML from the message content
-    String cleanedMessageContent = Jsoup.clean(messageContent, Whitelist.none());
+    if(type != null){
+      if(!conversation.getType().equals(type)){
+        conversation.makePrivate();
+        conversationStore.updateConversation(conversation);
+        System.out.println("Done!");
+      }
+    }
+    else if(messageContent != null) {
 
-    Message message =
-        new Message(
-            UUID.randomUUID(),
-            conversation.getId(),
-            user.getId(),
-            cleanedMessageContent,
-            Instant.now());
+      // this removes any HTML from the message content
+      String cleanedMessageContent = Jsoup.clean(messageContent, Whitelist.none());
 
-    messageStore.addMessage(message);
+      Message message =
+          new Message(
+              UUID.randomUUID(),
+              conversation.getId(),
+              user.getId(),
+              cleanedMessageContent,
+              Instant.now());
+
+      messageStore.addMessage(message);
+    }
 
     // redirect to a GET request
     response.sendRedirect("/chat/" + conversationTitle);
--- a/src/main/java/codeu/controller/ChatServlet.java
+++ b/src/main/java/codeu/controller/ChatServlet.java
@@ -101,6 +101,7 @@ public class ChatServlet extends HttpServlet {
 
     request.setAttribute("conversation", conversation);
     request.setAttribute("messages", messages);
+    request.setAttribute("isPrivate", conversation.isPrivate());
     request.getRequestDispatcher("/WEB-INF/view/chat.jsp").forward(request, response);
   }
 
@@ -144,11 +145,13 @@ public class ChatServlet extends HttpServlet {
     String messageContent = request.getParameter("message");
 
     if(type != null){
-      if(!conversation.getType().equals(type)){
+      if(type.equals("make private")){
         conversation.makePrivate();
-        conversationStore.updateConversation(conversation);
-        System.out.println("Done!");
       }
+      else{
+        conversation.makePublic();
+      }
+      conversationStore.updateConversation(conversation);
     }
     else if(messageContent != null) {
 
--- a/src/main/java/codeu/model/data/Conversation.java
+++ b/src/main/java/codeu/model/data/Conversation.java
@@ -26,7 +26,7 @@ public class Conversation {
   public final UUID owner;
   public final Instant creation;
   public final String title;
-  private String type;
+  private Boolean isPrivate;
 
   /**
    * Constructs a new Conversation.
@@ -41,7 +41,7 @@ public class Conversation {
     this.owner = owner;
     this.creation = creation;
     this.title = title;
-    this.type = "public";
+    this.isPrivate = false;
   }
 
   /** Returns the ID of this Conversation. */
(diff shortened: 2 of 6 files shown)


src/main/java/codeu/controller/ChatServlet.java
```diff
@@ -138,20 +140,31 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+      String cleanedMessageContent = Jsoup.clean(messageContent, Whitelist.none());
 
-    Message message =
-        new Message(
-            UUID.randomUUID(),
-            conversation.getId(),
-            user.getId(),
-            cleanedMessageContent,
-            Instant.now());
+      Message message =
```
lealgamb: spacing

src/main/webapp/WEB-INF/view/chat.jsp
```diff
@@ -28,6 +28,7 @@ String user = (String) request.getSession().getAttribute("user");
 <head>
   <title><%= conversation.getTitle() %></title>
   <link rel="stylesheet" href="/css/main.css" type="text/css">
+  <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
```
lealgamb: Can you explain what this is? Is it a font style?

src/test/java/codeu/controller/ChatServletTest.java
```diff
@@ -211,4 +214,52 @@ public void testDoPost_CleansHtmlContent() throws IOException, ServletException
...
+
+  @Test
+  public void testDoPost_makePrivatePrivate() throws IOException, ServletException {
+    Mockito.when(mockRequest.getRequestURI()).thenReturn("/chat/test_conversation");
+    Mockito.when(mockSession.getAttribute("user")).thenReturn("test_username");
+    Mockito.when(mockUserStore.getUser("test_username")).thenReturn(
+        new User(
+          UUID.randomUUID(),
+          "test_username",
+          "testHash",
```
lealgamb: nitpick: `test_hash` for consistency

src/test/java/codeu/controller/ChatServletTest.java
```diff
@@ -211,4 +214,52 @@ public void testDoPost_CleansHtmlContent() throws IOException, ServletException
...
+    
+    chatServlet.doPost(mockRequest, mockResponse);
+
+    Mockito.verify(mockConversation).makePrivate();
+    Mockito.verify(mockConversationStore).updateConversation(mockConversation);
+    Mockito.verify(mockResponse).sendRedirect("/chat/test_conversation");
+  }
+
+  @Test
+  public void testDoPost_makePrivatePrivate() throws IOException, ServletException {
```
lealgamb: The name of this test can be a little confusing because the test makes sure a private isn't made private.

src/main/java/codeu/model/data/Conversation.java
```diff
@@ -26,6 +26,7 @@
   public final UUID owner;
   public final Instant creation;
   public final String title;
+  private String type;
```
yebrahim: This is the sort of thing where a boolean makes things safer. You could easily mistype the value of the string field in other places in the code. How about making this `private Boolean isPrivate;`?

src/main/java/codeu/controller/ChatServlet.java
```diff
@@ -138,20 +140,31 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
     }
 
+    String type = request.getParameter("type");
     String messageContent = request.getParameter("message");
 
-    // this removes any HTML from the message content
-    String cleanedMessageContent = Jsoup.clean(messageContent, Whitelist.none());
+    if(type != null){
+      if(!conversation.getType().equals(type)){
+        conversation.makePrivate();
```
yebrahim: So it's a one-directional setting? Are you planning to allow users to switch a conversation back to being public?

src/main/webapp/WEB-INF/view/chat.jsp
```diff
@@ -43,6 +44,16 @@ String user = (String) request.getSession().getAttribute("user");
       var chatDiv = document.getElementById('chat');
       chatDiv.scrollTop = chatDiv.scrollHeight;
     };
+
+    // show contents of dropDown menu
+    function showDrop() {
```
yebrahim: I'd make the name clearer since it works with a specific dropdown element not any dropdown.

yebrahim: Also, since this switches the dropdown on and off, I'd maybe call it `toggleSettingsDropdown`?

src/main/webapp/WEB-INF/view/chat.jsp
```diff
@@ -43,6 +44,16 @@ String user = (String) request.getSession().getAttribute("user");
...
+      var dropDiv = document.getElementById("SettingDropDown");
+      if(dropDiv.style.display === "none") {
```
yebrahim: nit: tertiary operators make sense here:
`dropDiv.style.display = dropDiv.style.display === "none" ? "block" : "none";`

It's a matter of personal style though. :)

src/test/java/codeu/controller/ChatServletTest.java
```diff
@@ -211,4 +214,52 @@ public void testDoPost_CleansHtmlContent() throws IOException, ServletException
...
+    Mockito.when(mockConversationStore.getConversationWithTitle("test_conversation"))
+        .thenReturn(mockConversation);
+
+    Mockito.when(mockRequest.getParameter("type"))
+        .thenReturn("private");
+    Mockito.when(mockConversation.getType()).thenReturn("private");
+    
+    chatServlet.doPost(mockRequest, mockResponse);
+
+    Mockito.verify(mockConversation, Mockito.never()).makePrivate();
```
yebrahim: Perhaps we should just test that the conversation object isn't public at the end of the test, rather than verifying the mocks have/haven't been called?

src/main/java/codeu/controller/ChatServlet.java
```diff
@@ -138,20 +140,31 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+        conversationStore.updateConversation(conversation);
+        System.out.println("Done!");
```
yebrahim: If the purpose of the log is debugging, maybe the message should be a little clearer? Something like "Conversation <convo_name> switched to private by <user_id>"?

src/main/java/codeu/model/data/Conversation.java
jasminechau2: I agree that a boolean would be safer

src/main/webapp/WEB-INF/view/chat.jsp
vocatwk: This is for the gear icon

src/test/java/codeu/controller/ChatServletTest.java
vocatwk: lol I was blanking on names, all the others I could think of were very long. Any suggestions?

src/main/java/codeu/model/data/Conversation.java
vocatwk: Yes, you're right. will change it.

src/main/java/codeu/controller/ChatServlet.java
vocatwk: I was planning on implementing that too. I will add it to this PR.

src/test/java/codeu/controller/ChatServletTest.java
vocatwk: My goal was to test whether or not private conversation would be updated redundantly when a users clicks the private option on an already private conversation. is that not a relevant case to have a test for? and sorry about the test name, it could've been much clearer.

src/main/java/codeu/controller/ChatServlet.java
vocatwk: Oops, I had forgotten to remove that one.

src/test/java/codeu/controller/ChatServletTest.java
vocatwk: Although now that I'm implementing private -> publc, I will change up the tests a bit

src/test/java/codeu/controller/ChatServletTest.java
vocatwk: Actually nvm. I changed what the test is testing.

vocatwk:
The chat page now shows "make private"/"make public" option in the drop down depending on the current state of the conversation. Right now, any logged in user can use those buttons to change public conversations to private and vice versa but we can change that to allow admins or members depending on privacy settings that we could discuss.

I changed the second doPost test to test if it the chatServlet correctly changes private conversations to public.

src/main/java/codeu/controller/ChatServlet.java
```diff
@@ -144,11 +145,13 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
     String messageContent = request.getParameter("message");
 
     if(type != null){
-      if(!conversation.getType().equals(type)){
+      if(type.equals("make private")){
```
yebrahim: If we switch to using a toggle button (e.g. `<input type="checkbox" value="isPrivate">`), we can use `isPrivate` as a form attribute all over, instead of having to send "type" and converting it.

src/main/java/codeu/model/data/Conversation.java
```diff
@@ -26,6 +26,7 @@
   public final UUID owner;
   public final Instant creation;
   public final String title;
+  private Boolean isPrivate;
```
lealgamb: Just curious why we want the wrapper class for boolean?

src/main/java/codeu/controller/ChatServlet.java
vocatwk: So the users will see a value named "isPrivate" and we showed it checked or not depending on the state of the conversation?

src/main/java/codeu/model/data/Conversation.java
vocatwk: Didn't think about it. Guess it's not necessary.

src/main/java/codeu/controller/ChatServlet.java
yebrahim: Yes, and it works the other way too. Clicking the checkbox will toggle the type. It's up to you of course, I was just offering an alternative experience.

PR #46, opened 2018-06-08
Updated Event Implementation and Added Tests

lealgamb:
I'm now appending in the `EventStore` list and in the `jsp` file I am traversing from `[size - 1, 0]`. I added `getEvent()` based on a given `id`. 

An "About Me" update is now an event. 

Lastly, I wrote tests for `Event` and `EventStore` under the `model` folder. I will work on finishing writing tests under the `controller` folder.

image

### DIFF
--- a/src/test/java/codeu/model/store/basic/EventStoreTest.java
+++ b/src/test/java/codeu/model/store/basic/EventStoreTest.java
@@ -66,11 +66,10 @@ public class EventStoreTest {
   public void testAddEvent() {
     UUID eventId = UUID.randomUUID();
     List<String> testInformation = 
-        new ArrayList<>(
-          Arrays.asList(
-            "test_username", 
-            "test_conversation_title", 
-            "test_message_content"));
+        Arrays.asList(
+          "test_username", 
+          "test_conversation_title", 
+          "test_message_content");
     Event inputEvent = new Event(eventId, "Message", Instant.now(), testInformation);
 
     eventStore.addEvent(inputEvent);


src/test/java/codeu/model/store/basic/EventStoreTest.java
```diff
@@ -0,0 +1,89 @@
...
+    Event inputEvent = new Event(eventId, "Message", Instant.now(), testInformation);
+
+    eventStore.addEvent(inputEvent);
+    Event resultEvent = eventStore.getEvent(eventId);
+
+    assertEquals(inputEvent, resultEvent);
+    Mockito.verify(mockPersistentStorageAgent).writeThrough(inputEvent);
+  }
+
+  private void assertEquals(Event expectedEvent, Event actualEvent) {
```
vocatwk: Why not have an equals method in the event class if you're going to make one here anyway. Maybe it could be helpful somehow in the future. and this way you can also use it when comparing those methods in PersistentDataStoreTest.java

src/test/java/codeu/model/store/persistence/PersistentDataStoreTest.java
```diff
@@ -146,4 +148,77 @@ public void testSaveAndLoadMessages() throws PersistentDataStoreException {
...
+    persistentDataStore.writeThrough(aboutMeEvent);
+    persistentDataStore.writeThrough(conversationEvent);
+    persistentDataStore.writeThrough(messageEvent);
+
+    // load
+    List<Event> resultEvents = persistentDataStore.loadEvents();
+
+    // confirm that what we saved matches what we loaded
+    Event resultEventOne = resultEvents.get(0);
+    Assert.assertEquals(idOne, resultEventOne.getId());
```
vocatwk: You could instead use a method like before to compare resultEventOne and userEvent and the others below as well.

src/test/java/codeu/model/store/basic/EventStoreTest.java
```diff
@@ -0,0 +1,89 @@
...
+            "test_conversation_title", 
+            "test_message_content"));
+    Event inputEvent = new Event(eventId, "Message", Instant.now(), testInformation);
+
+    eventStore.addEvent(inputEvent);
+    Event resultEvent = eventStore.getEvent(eventId);
+
+    assertEquals(inputEvent, resultEvent);
+    Mockito.verify(mockPersistentStorageAgent).writeThrough(inputEvent);
+  }
```
yebrahim: How about a test to ensure events are retrieved in order? We rely on this behavior in the JSP to show them correctly right?

src/test/java/codeu/model/store/persistence/PersistentDataStoreTest.java
```diff
@@ -146,4 +148,77 @@ public void testSaveAndLoadMessages() throws PersistentDataStoreException {
     Assert.assertEquals(contentTwo, resultMessageTwo.getContent());
     Assert.assertEquals(creationTwo, resultMessageTwo.getCreationTime());
   }
+
+  @Test
+  public void testSaveAndLoadEvents() throws PersistentDataStoreException {
+    UUID idOne = UUID.randomUUID();
```
yebrahim: Can you use an array and a for loop to generate this test data? Something like (didn't check syntax, probably incorrect):
```
List<String> eventTypes = Arrays.asList("User", "About Me", "Conversation", "Message");
List<Event> events = new ArrayList<Event>();
// Generate test events
for (int 0; i < 4; ++i) {
  events.add(new Event(
    UUID.randomUUID(),
    eventTypes[i],
    Instant.ofEpochMilli(1000 * (i + 1)),
    Arrays.asList("user_" + i, "some more information for event #" + i)
  ));
}
// Write test events
for (Event e: events) {
  persistentDataStore.writeThrough(e);
}
```

src/test/java/codeu/model/store/basic/EventStoreTest.java
lealgamb: I added it here to be consistent with the other test files that were in the original code base? Do you guys think I should still move this function to the Event class?

src/test/java/codeu/model/store/persistence/PersistentDataStoreTest.java
lealgamb: I also did this to be consistent with the other tests. Do you think I should change it?

src/main/java/codeu/model/store/persistence/PersistentDataStore.java
```diff
@@ -164,7 +164,7 @@ public PersistentDataStore() {
     List<Event> events = new ArrayList<>();
 
     // Retrieve all events from the datastore.
-    Query query = new Query("chat-events").addSort("creation_time", SortDirection.DESCENDING);
+    Query query = new Query("chat-events").addSort("creation_time", SortDirection.ASCENDING);
```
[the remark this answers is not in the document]
lealgamb: I believe that the events will always be in order, but I based this off of the original code. Should I remove it?

src/test/java/codeu/model/store/basic/EventStoreTest.java
lealgamb: Yes, I'll work on this.

src/test/java/codeu/model/store/persistence/PersistentDataStoreTest.java
lealgamb: I'll update this, it's much cleaner.

src/main/java/codeu/model/store/persistence/PersistentDataStore.java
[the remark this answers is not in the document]
yebrahim: I think we should only keep it if we know that it adds something. Otherwise, let's at least remove it and see if it breaks our tests, which should make sure events are sorted.

lealgamb:
I updated the `controller` tests.

src/main/java/codeu/model/store/persistence/PersistentDataStore.java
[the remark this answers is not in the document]
lealgamb: The test that saves and loads events fails after removing this.

src/test/java/codeu/model/store/basic/EventStoreTest.java
vocatwk: Well if it were, you could use it in the PersistentDataStoreTest.java as well to compare the events instead of a for loop comparing each property of the event. But it's not that big of deal, maybe it is better to keep it consistent with the others.

src/test/java/codeu/controller/ActivityFeedServletTest.java
```diff
@@ -39,47 +36,48 @@ public void setUp() {
...
+    mockEventStore = Mockito.mock(EventStore.class);
+    activityFeedServlet.setEventStore(mockEventStore);
   }
 
   @Test
   public void testDoGet() throws IOException, ServletException {
+    List<Event> fakeEventList = new ArrayList<>();
+
     UUID fakeUserId = UUID.randomUUID();
     User fakeUser =
```
vocatwk: Not really sure if you need a fake user since you can just manually fill in dummy values to create the event object.

src/test/java/codeu/controller/ActivityFeedServletTest.java
```diff
@@ -39,47 +36,48 @@ public void setUp() {
...
-    List<Message> fakeMessageList = new ArrayList<>();
-    fakeMessageList.add(
-          new Message(
-            UUID.randomUUID(), 
-            fakeConversationId,
-            fakeUserId,
-            "test_message",
-            Instant.now()));
-    Mockito.when(mockMessageStore.getAllMessages()).thenReturn(fakeMessageList);
+        new Conversation(fakeConversationId, fakeUserId, "test_conversation", Instant.now());
```
vocatwk: same as above. not sure if a conversation object is needed.

src/test/java/codeu/controller/ActivityFeedServletTest.java
```diff
@@ -39,47 +36,48 @@ public void setUp() {
...
+    List<String> conversationInformation = new ArrayList<>();
+    conversationInformation.add(fakeUser.getName());
+    conversationInformation.add(fakeConversation.getTitle());
+    Event conversationEvent = new Event(
+        UUID.randomUUID(), "Conversation", fakeConversation.getCreationTime(), conversationInformation);
+    fakeEventList.add(conversationEvent);
+    
+    Message fakeMessage = new Message(
```
vocatwk: same as above.

src/test/java/codeu/controller/ProfileServletTest.java
```diff
@@ -112,6 +120,22 @@ public void testDoPost_SelfProfile() throws IOException, ServletException {
 
     Mockito.verify(mockUser).setAboutMe("someMessage");
     Mockito.verify(mockUserStore).updateUser(mockUser);
+
+    ArgumentCaptor<User> userArgumentCaptor =
```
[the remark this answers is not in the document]
lealgamb: You're right, I'll remove those in the next PR so I can merge this one.

PR #55, opened 2018-06-17
Added a search bar, only searching Users for now

lealgamb:
Added a search bar to the navigation bar. Empty searches are disabled. Currently only searching for users and getting/displaying them as a list. Will add tests in another PR.

Search Bar:
image

Empty Search:
image

Valid Search:
image

### DIFF
--- a/src/main/java/codeu/controller/SearchServlet.java
+++ b/src/main/java/codeu/controller/SearchServlet.java
@@ -8,7 +8,6 @@ import codeu.model.store.basic.ConversationStore;
 import codeu.model.store.basic.MessageStore;
 import java.io.IOException;
 import java.util.List;
-import java.util.ArrayList;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
--- a/src/main/java/codeu/controller/SearchServlet.java
+++ b/src/main/java/codeu/controller/SearchServlet.java
@@ -1,0 +1,92 @@
+package codeu.controller;
+
+import codeu.model.data.User;
+import codeu.model.data.Conversation;
+import codeu.model.data.Message;
+import codeu.model.store.basic.UserStore;
+import codeu.model.store.basic.ConversationStore;
+import codeu.model.store.basic.MessageStore;
+import java.io.IOException;
+import java.util.List;
+import java.util.ArrayList;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet class responsible for the search feature. */
+public class SearchServlet extends HttpServlet {
+
+  /** Store class that gives access to Users. */
+  private UserStore userStore;
+
+  /** Store class that gives access to Conversations. */
+  private ConversationStore conversationStore;
+
+  /** Store class that gives access to Messages. */
+  private MessageStore messageStore;
+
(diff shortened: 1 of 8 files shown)


jasminechau2:
It LGTM!

src/main/webapp/css/main.css
```diff
@@ -35,6 +35,13 @@ nav a {
   text-decoration: none;
 }
 
+nav form {
```
vocatwk: Sorry couldn't find where you used this.

vocatwk:
Not sure what you guys prefer but could we also have the search results displayed on top of a page as like a drop down instead of a different page?

src/main/webapp/WEB-INF/view/search.jsp
```diff
@@ -0,0 +1,44 @@
...
+        <p>No results for: <%= searchRequest %></p>
+      <%
+      }
+      else {
+      %>
+        <p>Results for: <%= searchRequest %></p>
+      <%
+        for (User user : users) {
+      %>
+          <p><%= user.getName() %></p>
```
yebrahim: How about we make this a link to the user's profile page?

src/main/java/codeu/model/store/basic/UserStore.java
```diff
@@ -98,6 +98,27 @@ public User getUser(UUID id) {
...
   }
 
+  /**
+  * Access the Users that contains the given username.
+  * 
+  * Used with search requests. 
+  * 
+  * @return null if username does not match any existing user
+  */
+  public List<User> getUsers(String username) {
```
yebrahim: Tests for this method?

src/main/java/codeu/controller/SearchServlet.java
```diff
@@ -0,0 +1,91 @@
+package codeu.controller;
```
yebrahim: Tests for this servlet?

yebrahim:
Nice work on the UI too, the search box looks good.

src/main/webapp/css/main.css
lealgamb: It's for the form that the search bar is contained in.

src/main/webapp/WEB-INF/view/search.jsp
lealgamb: Will do, forgot to mention in the description that this was only to test the search function.

src/main/java/codeu/controller/SearchServlet.java
lealgamb: Will add the tests in a future PR, sorry I forgot to mention that.

lealgamb:
@vocatwk I will add that feature either in this PR or a new one. Are you thinking of a dynamic drop down menu, like it searches as you type? And if the user clicks the button then it redirects to the other page?

lealgamb:
I'm having some trouble figuring out how to implement the instant search. Any tips?

vocatwk:
@lealgamb that sounds good to me!

vocatwk:
is there a way to detect if the value of search bar changes and update your search results each time it does. and you could preset the value to the value you just used to search for data so that you don't interrupt the user's search. It sounds very expensive but can't think about any other way of doing it at the moment.

yebrahim:
Looks good, but can we merge the PR if it’s ready with the separate results page, then work on the dynamic drop down in a separate PR?

yebrahim:
After merging this, can we add a little section in the PR about search? Just a sentence or two about the fact that we offer searching for usernames right from the top bar.

lealgamb:
Yup, will merge this now.

PR #57, opened 2018-06-18
Jasmine notifications

jasminechau2:
I haven't added any code to the persistent data store, I wanted to get a review before I did all that to see if I am on a reasonable track. Could there be a way of doing this without creating a notification store?


yebrahim:
Looks like this branch includes some of the changes in the admin branch, you probably branched off of that instead of master. :)
Do these changes require some of the changes in the admin branch in order to work, or are they completely separate?

lealgamb:
The notification store might come in handy. There might also be a way to use the event store and adding a "last seen" timestamp like it was mentioned in an older PR. For example, if there's a drop down notification menu (similar to Facebook), you can just get all events with a timestamp that are greater than the "last seen" timestamp.

jasminechau2:
The admin changes don't do anything the to affect the notification page

src/main/java/codeu/model/data/Notification.java
```diff
@@ -0,0 +1,59 @@
...
+
+  public User getSender(){
+    return sender;
+  }
+
+  public UUID getId(){
+    return id;
+  }
+
+  public List<String> getNotification(){
```
vocatwk: is there a reason you need these as a list? Maybe you could have getSender(), getReceiver() ... methods that will get that information for you from the notification object right.

vocatwk:
Hmm.. Maybe you could use the EventStore as your NotificationStore depending on how you implement it. I really like using the last seen timestamp.  Also, what events are we going to have notifications for, messages? users joining? ... it looks like right now the Notifications class is suited for messages.

yebrahim:
I'd rather review the PR after you rebase it against master, or create a new branch off of master and re-apply your notifications-related commits. Can you please take care of this first? :)

PR #59, opened 2018-06-19
change branches to make reviews clearer

jasminechau2:
Notification System: have not added it to the persistent data store.


yebrahim:
Thanks, this is much cleaner and easier to review now.

Can we explore the alternative suggested by @lealgamb on the other PR? Namely, to just store an extra field `lastSeenTimestamp` or something into the `User` class, which we update whenever the user logs in, and we use to calculate which events to show to users? This is a simple implementation I think, and to me it follows from the implementation of the `Event` class.

@jasminechau2 What do you think?

jasminechau2:
I can do that @yebrahim

lealgamb:
Just wondering if this is ready for a review.

jasminechau2:
You can It's a bit messy since I was updating the wrong branch and had to transfer code over @yebrahim

yebrahim:
You can leave this PR open to borrow code from it later, and start a new one with just the skeleton for the prototype, this should make things easier. :)

PR #66, opened 2018-06-25
Updated Notifications MVP

jasminechau2:
Have yet to implement a way to see notifications but provides an empty page for the information to go to.

### DIFF
--- a/src/main/java/codeu/controller/NotificationServlet.java
+++ b/src/main/java/codeu/controller/NotificationServlet.java
@@ -1,0 +1,57 @@
+package codeu.controller;
+
+import codeu.model.data.User;
+import codeu.model.data.Notification;
+import codeu.model.store.basic.UserStore;
+import codeu.model.store.basic.NotificationStore;
+import java.io.IOException;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.util.ArrayList;
+import java.util.List;
+import org.mindrot.jbcrypt.BCrypt;
+import java.util.UUID;
+import java.util.Arrays;
+
+public class NotificationServlet extends HttpServlet{
+  private UserStore userStore;
+  private NotificationStore notificationStore;
+
+  @Override
+  public void init() throws ServletException {
+    super.init();
+    setUserStore(UserStore.getInstance());
+    setNotificationStore(NotificationStore.getInstance());
+  }
+  /**
+   * Sets the UserStore used by this servlet. This function provides a common setup method for use
+   * by the test framework or the servlet's init() function.
+   */
+  void setUserStore(UserStore userStore) {
+    this.userStore = userStore;
+  }
+
+  /**
+   * Sets the NotificationStore used by this servlet. This function provides a common setup method for use
+   * by the test framework or the servlet's init() function.
+   */
+  void setNotificationStore(NotificationStore notificationStore){
+    this.notificationStore = notificationStore;
+  }
+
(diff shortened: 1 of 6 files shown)


yebrahim:
I think this PR is still adding a lot of details that we still need to discuss.
How about reducing it to just the prototype? Just a servlet that sends a JSP file that has a hello world or something?
We can then discuss what makes a notification, and how to get a list for each user.

jasminechau2:
Simplified it @yebrahim

src/main/java/codeu/controller/NotificationServlet.java
```diff
@@ -0,0 +1,47 @@
+package codeu.controller;
+
+import codeu.model.data.User;
+import codeu.model.store.basic.UserStore;
+import java.io.IOException;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.util.ArrayList;
```
yebrahim: I think some of these aren't used, let's scrub them to keep only what we need in this PR.

src/main/java/codeu/controller/NotificationServlet.java
```diff
@@ -0,0 +1,47 @@
...
+  }
+  /**
+   * Sets the UserStore used by this servlet. This function provides a common setup method for use
+   * by the test framework or the servlet's init() function.
+   */
+  void setUserStore(UserStore userStore) {
+    this.userStore = userStore;
+  }
+
+  /**
```
yebrahim: Remove this comment block?

src/main/java/codeu/controller/NotificationServlet.java
```diff
@@ -0,0 +1,47 @@
...
+   * Sets the NotificationStore used by this servlet. This function provides a common setup method for use
+   * by the test framework or the servlet's init() function.
+   */
+
+//TODO Decide if other notification types beyond messages are needed
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response)
+      throws IOException, ServletException {
+        String username = (String)request.getSession().getAttribute("user");
```
yebrahim: We're not doing anything in these lines, can we just keep the request forwarding line and remove the rest?

src/main/webapp/WEB-INF/view/notification.jsp
```diff
@@ -0,0 +1,17 @@
+<%@ page import="codeu.model.data.User" %>
```
yebrahim: Remove unused imports here?

src/main/webapp/WEB-INF/view/navbar.jsp
```diff
@@ -16,6 +16,7 @@
   <% } %>
   <a href="/about.jsp">About</a>
   <a href="/activityfeed">Activity Feed</a>
+  <a href="/notifcation"> Notifications</a>
```
yebrahim: Can we rename the path to `/notifications`, just like how we call the link?

src/main/java/codeu/controller/NotificationServlet.java
```diff
@@ -0,0 +1,35 @@
...
+   */
+  void setUserStore(UserStore userStore) {
+    this.userStore = userStore;
+  }
+
+//TODO Decide if other notification types beyond messages are needed
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response)
+      throws IOException, ServletException {
+        request.getRequestDispatcher("/WEB-INF/view/notification.jsp").forward(request, response);
```
yebrahim: Should we also change the JSP file name and reference here? :)

Issue #64, opened 2018-06-21
Complete Dynamic Dropdown Search


lealgamb:

yebrahim:
You can add "Fixes #64" to the outstanding PR so it closes this one automatically.

PR #67, opened 2018-06-26
Dynamic Search Bar - Users

### LINKED ISSUE #62: Search Bar Instant Search - Using Fetch Request
I've been working on the steps @yebrahim provided and this is what I have so far. 

In `SearchServlet` I added the following in `doGet()`:

```
String json = new Gson().toJson(users);
response.setContentType("application/json");
response.setCharacterEncoding("UTF-8");
response.getWriter().write(json);
```

This successfully created a JSON object, but it prints it on the page:
image

My questions are:

1. How do I only make the JSON object have the following:
```
[{
  "name": "user1",
}, {
  "name": "user2"
}]
```

2. I understand how to call a function using keypress, but how do I call the `[fetch]` function?

lealgamb:
The instant search feature works now for Users. Currently only 3 results are displayed and I added a scrollbar as needed.

Here is a problem I see for adding Conversations and Messages:

- To parse the JSON objects, it might be a good idea to add `getLink()` to the 3 classes to be able to directly access the link when parsing instead of determining the type and harcoding the link.

Other problems I see are:

- How to prevent the user from accessing `/search` which has the JSON object printed.

- Related to this, should the form still have the `action` and `method` attributes?

- Allowing only alphanumeric characters in search. Currently a search with multiple spaces returns all of the users

This is more of a design issue, but it would be good to keep in mind. Basically, have the results overlap the nav bar so that the nav bar itself doesn't "grow" as the results come in.

image

Fixes #62, #64


src/main/java/codeu/controller/SearchServlet.java
```diff
@@ -65,18 +66,24 @@ void setMessageStore(MessageStore messageStore) {
   public void doGet(HttpServletRequest request, HttpServletResponse response) 
       throws IOException, ServletException {
     String searchRequest = (String)request.getParameter("searchRequest");
-    request.setAttribute("searchRequest", searchRequest);
 
+    // TODO 
+    if (searchRequest == null) {
```
yebrahim: You can take out the empty if condition, and add some more meat to the TODO?

src/main/java/codeu/controller/SearchServlet.java
```diff
@@ -65,18 +66,24 @@ void setMessageStore(MessageStore messageStore) {
...
+    
     List<User> users = userStore.getUsers(searchRequest);
-    request.setAttribute("users", users);
+    
+    String json = new Gson().toJson(users);
+    response.setContentType("application/json");
+    response.setCharacterEncoding("UTF-8");
+    response.getWriter().write(json);
 
     // List<Conversation> conversations = conversationStore.getAllConversations();
```
yebrahim: Can we clean up these comments? If there's anything you want to remember to do, add a `// TODO` line instead.

src/main/webapp/WEB-INF/view/searchbar.jsp
```diff
@@ -16,30 +9,63 @@ List<User> allUsers = UserStore.getInstance().getAllUsers();
...
   }
 
-  // Toggle drop down.
-  function toggleDropDown() {
-    var searchBar = document.getElementById("searchBar");
-    var dropDiv = document.getElementById("dropDiv");
-    if (searchBar.value == "") {
-      dropDiv.style.display = "none";
+  // This is to remove the resultItem divs
+  function removeElementsByClass(className){
```
yebrahim: Alternatively, you can just set the innerHTML attribute of the result div to an empty string, this removes all children:
`document.querySelector('#result').innerHTML = '';`
This way you don't need the classes (or the IDs) on the result divs.

src/main/webapp/WEB-INF/view/searchbar.jsp
```diff
@@ -16,30 +9,63 @@ List<User> allUsers = UserStore.getInstance().getAllUsers();
...
+      .then(
+        function(response) {
+          if (response.status !== 200) {
+            console.log("Looks like there was a problem. Status Code: " + response.status);
+            return;
+          }
+          response.json().then(function(data) {
+            if (data != null) {
+              // Start with no resultItem divs
+              removeElementsByClass("resultItem");
```
yebrahim: We're always going to want to clear out the results divs right? It doesn't matter if we have data or not.

src/main/webapp/WEB-INF/view/searchbar.jsp
```diff
@@ -16,30 +9,63 @@ List<User> allUsers = UserStore.getInstance().getAllUsers();
...
+              for (var user in data) {
+                var userName = data[user].name;
+                var div = document.createElement("div");
+                div.setAttribute("class", "resultItem");
+                div.setAttribute("id", userName);
```
yebrahim: What're we using the id for?

src/main/webapp/WEB-INF/view/searchbar.jsp
```diff
@@ -16,30 +9,63 @@ List<User> allUsers = UserStore.getInstance().getAllUsers();
...
+        }
+      ) 
+      .catch(function(err) {
+        console.log("Fetch Error :-S", err);
+      });
     }
+    // Clear the results when the search bar is empty
     else {
-      dropDiv.style.display = "block"
+      removeElementsByClass("resultItem");
```
yebrahim: Here too, we're always going to clear out the results on each keypress, how about we do this once at the beginning of the function, and get rid of the if conditions?

src/main/webapp/css/main.css
```diff
@@ -78,9 +82,36 @@ button {
   overflow-y: scroll;
 }
 
+/*The following will probably not be needed.*/
```
yebrahim: Let's just remove all the commented out code. If we need it back, we can get it from source control. :)

src/main/webapp/css/main.css
```diff
@@ -78,9 +82,36 @@ button {
...
 /* This is for the Search Results Page */
-#results {
+/*#results {
   background-color: white;
   height: 500px;
   overflow-y: scroll;
+}*/
+
+#result {
+  background-color: #eeeeee;
```
yebrahim: Try adding `position: absolute;` here, and maybe a minimum width or something.

src/main/webapp/css/main.css
```diff
@@ -78,9 +82,36 @@ button {
...
+  font-size: 16px;
+  color: black;
+  margin: 0;
+  padding: 0;
+  display: block;
+  height: 100%;
+  width: 100%;
+}
+
+.resultItem a:hover {
```
yebrahim: Neat!

src/main/webapp/WEB-INF/view/searchbar.jsp
vocatwk: wow, the secrets of javascript. 😂

src/main/webapp/WEB-INF/view/searchbar.jsp
```diff
@@ -16,30 +9,63 @@ List<User> allUsers = UserStore.getInstance().getAllUsers();
...
-    <%
-    for (User u : allUsers) {
-    String username = u.getName();
-    %>
-      <a href="/profile/<%= username %>"><%= username %></a>
-      <br>
-    <%
-    }
-    %>
+  <input onkeyup="fetchResults()" type="text" autocomplete="off" name="searchRequest" id="searchBar"><button type="submit" onclick="return isEmpty();"><i class="fa fa-search"></i></button>
```
vocatwk: What is supposed to happen when they click the search button?

vocatwk:
Sorry, I'm a bit confused about what you meant by your first comment. What is the link you're referring to?

src/main/webapp/WEB-INF/view/searchbar.jsp
lealgamb: I was using it to debug and forgot to remove it. Will clean it up.

src/main/webapp/WEB-INF/view/searchbar.jsp
lealgamb: That's also what I'm trying to figure out. Since it's an instant search now I think it can be removed. Is that okay with you guys?

src/main/webapp/WEB-INF/view/searchbar.jsp
lealgamb: Thanks for this tip, it makes the code a lot cleaner.

lealgamb:
If resultItems are cleared on every keypress, then it makes the animation kind of glitchy. It's like it's "blinking" on every keypress.

lealgamb:
@vocatwk I'm talking about the links for Users and Conversations we have. (/profile/*, /chat/*). Since the search only works for Users then it's fine for now.

lealgamb:
When the users access `/search` there's a NPE. I believe it's because `searchRequest` is null. 
image

If the user provides a search request, something like `/search?searchRequest=leal`, then they will see the following.
image

And if they provide a search request that doesn't exist, then they will see this (which is correct).
image

That's why I'm wondering if there's a way to restrict users from directly accessing this page.

src/main/java/codeu/controller/SearchServlet.java
```diff
@@ -43,49 +32,22 @@ void setUserStore(UserStore userStore) {
...
+   * This function gets the Users that contains the searchRequest (username) 
+   * and sets up the JSON object to be parsed. 
    */
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) 
       throws IOException, ServletException {
     String searchRequest = (String)request.getParameter("searchRequest");
-    request.setAttribute("searchRequest", searchRequest);
 
+    // TODO: Determine what to do when searchRequest is null
```
yebrahim: Maybe just return an empty list? `json = '[]';`

src/main/webapp/WEB-INF/view/searchbar.jsp
```diff
@@ -1,45 +1,44 @@
...
+      fetch('/search?searchRequest=' + searchBarValue, 
+        {credentials: "same-origin"})
+      .then(
+        function(response) {
+          if (response.status !== 200) {
+            console.log("Looks like there was a problem. Status Code: " + response.status);
+            return;
+          }
+          response.json().then(function(data) {
+            if (data != null) {
```
yebrahim: nit: this `if` is redundant, the for loop below won't execute if data is null.

yebrahim:
Sorry, submitted the review before I saw your other comments. :)

- I'm fine if you want to keep the condition for clearing the results, if you don't like the flicker.
- I think it's fine to let users access `/search` from the browser, it's not really exposing any new data, and the user shouldn't know about its existence in the first place. If you do want to hide it, there are some tricks you can do, such as requiring a special header that you only send in the `fetch` call, such that if you don't find it in the servlet, you can just redirect the user to the index page, or not do anything at all.

src/main/java/codeu/controller/SearchServlet.java
lealgamb: Thanks, will do.

src/main/webapp/WEB-INF/view/searchbar.jsp
lealgamb: Thanks for pointing this out

lealgamb:
Ah, I hadn't realized that the user shouldn't know about `/search`. This simplifies things a lot, thanks! 

Also, I'm not a big fan of the flicker so I'll revert the changes if it's okay with everyone else.

jasminechau2:
It's okay by me

yebrahim:
Sure.

src/main/java/codeu/controller/SearchServlet.java
```diff
@@ -43,49 +32,26 @@ void setUserStore(UserStore userStore) {
...
-  /**
-  * TODO: Determine if this is needed.
-  */
-  @Override
-  public void doPost(HttpServletRequest request, HttpServletResponse response) 
-      throws IOException, ServletException {
-    // TODO
+    String json;
+    if (searchRequest == null) {
+      json = new Gson().toJson("[]");
```
yebrahim: Why not just `json = "[]";`?

src/main/java/codeu/controller/SearchServlet.java
```diff
@@ -43,49 +32,26 @@ void setUserStore(UserStore userStore) {
...
+    }
+    else {
+      List<User> users = userStore.getUsers(searchRequest);
```
yebrahim: If we make `getUsers` return an empty list when `searchRequest` is null, an empty string, or a non-matching string, then this simplifies this code block, it then just becomes:
`String json = new Gson().toJson(userStore.getUsers(searchRequest));`
It also makes sense that the return type of the function is consistent to always be a list.

src/main/webapp/WEB-INF/view/searchbar.jsp
```diff
@@ -1,45 +1,53 @@
...
+              }
+              // Data is null
+              else {
+                // Clear the resultItem divs
+                document.querySelector('#result').innerHTML = '';
+              }
+            });
+          }
+        ) 
+        .catch(function(err) {
```
yebrahim: nit: you could make use of the more compact lambda syntax. Basically, these two are (mostly*) equivalent:
```
catch(function(err) {
  console.log(...);
});
```
and:
```
catch(err => console.log(...));
```
----
\* there are some semantic differences.

lealgamb:
All good to merge?

yebrahim:
Lgtm.

lealgamb:
Yeah, no more flicker. 😃

PR #69, opened 2018-06-27
Adding a new method to the User class check if the user has visited the notification page

jasminechau2:
Creating Notification Class


src/main/java/codeu/model/data/User.java
```diff
@@ -40,6 +41,7 @@ public User(UUID id, String name, String passwordHash, Instant creation) {
     this.creation = creation;
     aboutMe = "";
     isAdmin = false;
+
```
lealgamb: It might be a good idea to set a value for `lastSeenNotificationsTimestamp` in the constructor. Perhaps it can be null that way we know the user is new/hasn't seen any notifications.

src/main/java/codeu/controller/NotificationServlet.java
```diff
@@ -8,6 +8,8 @@
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import org.mindrot.jbcrypt.BCrypt;
+import java.util.ArrayList;
```
lealgamb: nit: these will probably be used later on but it's good to only keep the imports that are needed for this PR

src/main/java/codeu/model/data/User.java
yebrahim: +1.
You can also add a unit test for it, to make sure it gets initialized to null for a new user.

yebrahim:
How about we change the PR title to reflect the changes? Something about adding the new field to the User class. :)

src/main/java/codeu/model/data/User.java
jasminechau2: Would it be better to set the time the creation of the user class? Instant has a function now() that sets the time at the current moment because we wouldn't want new users to see past notifications.

yebrahim: I think this is actually more accurate, thanks for catching that. Yes we should probably use the current time as value for this field on new users.

src/main/java/codeu/controller/NotificationServlet.java
```diff
@@ -2,6 +2,8 @@
 
 import codeu.model.data.User;
 import codeu.model.store.basic.UserStore;
+import codeu.model.data.Event;
```
yebrahim: I don’t think you need most of these changes. Only the initialization of the new field in User.java.

src/main/webapp/WEB-INF/view/notifications.jsp
```diff
@@ -14,5 +14,6 @@
 <body>
   <%@ include file="navbar.jsp" %>
   <p> Hello User This is your notifcation page</p>
+  <% request.getAttribute("eventsToShow");%>
```
vocatwk: Where do you set the eventsToShow attribute?

src/main/java/codeu/controller/NotificationServlet.java
```diff
@@ -8,9 +8,14 @@
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import org.mindrot.jbcrypt.BCrypt;
+import java.util.ArrayList;
+import java.util.List;
+import java.time.Instant;
 
 public class NotificationServlet extends HttpServlet{
   private UserStore userStore;
+  private Instant timeUserVistedPage;
```
yebrahim: Why do we need this?

PR #72, opened 2018-06-30
Search Bar Tests

### LINKED ISSUE #58: Search Bar Tests

lealgamb:
Added tests for the search bar. 

Fixes #58

### DIFF
--- a/src/test/java/codeu/controller/SearchServletTest.java
+++ b/src/test/java/codeu/controller/SearchServletTest.java
@@ -1,0 +1,62 @@
+package codeu.controller;
+
+import codeu.model.data.User;
+import codeu.model.store.basic.UserStore;
+import java.io.IOException;
+import java.io.StringWriter;
+import java.io.PrintWriter;
+import java.time.Instant;
+import java.util.UUID;
+import java.util.List;
+import java.util.ArrayList;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+import com.google.gson.Gson;
+
+public class SearchServletTest {
+
+  private SearchServlet searchServlet;
+  private HttpServletRequest mockRequest;
+  private HttpServletResponse mockResponse;
+  private UserStore mockUserStore;
+
+  @Before
+  public void setup() {
+    searchServlet = new SearchServlet();
+
+    mockRequest = Mockito.mock(HttpServletRequest.class);
+    mockResponse = Mockito.mock(HttpServletResponse.class);
+    
+    mockUserStore = Mockito.mock(UserStore.class);
+    searchServlet.setUserStore(mockUserStore);
+  }
+
+  @Test
+  public void testDoGet() throws IOException, ServletException {
+    List<User> fakeUserList = new ArrayList<User>();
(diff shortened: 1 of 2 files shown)


src/test/java/codeu/controller/SearchServletTest.java
```diff
@@ -0,0 +1,62 @@
...
+    StringWriter stringWriter = new StringWriter();
+    PrintWriter writer = new PrintWriter(stringWriter);
+    Mockito.when(mockResponse.getWriter()).thenReturn(writer);
+
+    searchServlet.doGet(mockRequest, mockResponse);
+
+    Mockito.when(mockResponse.getContentType()).thenReturn("application/json");
+    Mockito.when(mockResponse.getCharacterEncoding()).thenReturn("UTF-8");
+
+    Assert.assertEquals("application/json", mockResponse.getContentType());
```
lealgamb: I'm curious why lines 55-56 are needed for lines 58-59 to work? I still can't completely wrap my head around this.

vocatwk: I'm not sure exactly why you're calling getContentType() and getCharacterEncoding() on the response object. If you want to verify that it calls the setter methods correctly with the correct arguments, you could just use argument captor.

src/test/java/codeu/controller/SearchServletTest.java
```diff
@@ -0,0 +1,62 @@
...
+    Assert.assertEquals("UTF-8", mockResponse.getCharacterEncoding());
+    Assert.assertEquals(json, stringWriter.toString());
```
vocatwk: I'm also a little confused about this line. You created the json object and the stringWriter in this test method, what does this have to do with testing the doGet() method.

src/test/java/codeu/controller/SearchServletTest.java
yebrahim: It confused me for a while too, but once I found [this SO question](https://stackoverflow.com/questions/50403688/return-of-getcontenttype-in-mocked-httpservletresponse-instance-is-null) I thought, of course! You're not using an actual `HttpServletResponse` object here, so the `setContentType` call doesn't really do anything.

src/test/java/codeu/model/store/basic/UserStoreTest.java
```diff
@@ -75,6 +75,33 @@ public void testGetUser_byId_notFound() {
...
 
+  @Test
+  public void testGetUsers_byUsername_found() {
+    List<User> resultUsers = userStore.getUsers("username");
+    Assert.assertEquals(3, resultUsers.size());
+    assertEquals(USER_ONE, resultUsers.get(0));
+    assertEquals(USER_TWO, resultUsers.get(1));
+    assertEquals(USER_THREE, resultUsers.get(2));
+  }
+
```
yebrahim: How about one more test here to make sure partial strings are matched? For your sample users in this file, "USER_T" for example should return two results: "USER_TWO" and "USER_THREE".

src/test/java/codeu/controller/SearchServletTest.java
lealgamb: That link cleared things up for me, thanks. I tried using an argument captor but I understood the link example more.

src/test/java/codeu/controller/SearchServletTest.java
lealgamb: I included this to verify that the json is created properly.

PR #73, opened 2018-06-30
added add user option in conversation

vocatwk:
I added a placeholder box for the add user box when you click the add user button which is in the settings drop down. I also added a members variable to the conversation class and some related methods. I modified the tests that include the conversation class to reflect those changes (also ones that involved the user class because I forgot to do that last time.)

I was thinking of making the links in the drop down settings menu anchor tags so they look more uniform but since the make private button sends a post request, I have to use jquery to send the POST request. I wanted to ask your input on that first.

I modified Guillermo's code a little so I can reuse it for the search bar in the modal. Let me know if I should change it back or do it a different way.

Also I wanted to close the modal whenever a user clicks outside but I couldn't get it to work. I'll work on it in the mean time and add it if it works.

Moving forward, I will implement actually adding users part, then move onto displaying conversations only to members.

screen shot 2018-06-30 at 10 38 05 am
screen shot 2018-06-30 at 1 15 55 pm


src/main/java/codeu/model/store/persistence/PersistentDataStore.java
```diff
@@ -107,9 +108,11 @@ public PersistentDataStore() {
         String title = (String) entity.getProperty("title");
         Instant creationTime = Instant.parse((String) entity.getProperty("creation_time"));
         Boolean isPrivate = (Boolean) entity.getProperty("is_private");
+        List<String> membersAsList = (List<String>) entity.getProperty("members");
```
lealgamb: Is it possible to directly cast as a HashSet?

src/main/webapp/WEB-INF/view/addUserBox.jsp
```diff
@@ -0,0 +1,18 @@
+<!-- Add user modal -->
+<div id="addUserBox">
+
+  <!-- modal content -->
+  <div id="addUserContent">
+
+    <span class="close">&times;</span>
+    <p> Add people to this conversation! </p>
+    <div id ="searchDiv">
```
lealgamb: Might be a good idea to use a class instead of an id now that we have the search bar in multiple places.

src/main/webapp/WEB-INF/view/searchbar.jsp
```diff
@@ -1,7 +1,7 @@
 <script type="text/javascript">
   // Iterates through fetched results
-  function fetchResults() {
-    var searchBarValue = document.querySelector('#searchBar').value;
+  function fetchResults(putResultsIn, forValue) {
```
lealgamb: nice job extending this 👍

src/test/java/codeu/model/data/ConversationTest.java
```diff
@@ -27,12 +28,32 @@ public void testCreate() {
     UUID owner = UUID.randomUUID();
     String title = "Test_Title";
     Instant creation = Instant.now();
+    Boolean isPrivate = true;
+    HashSet<String> members = new HashSet<>();
```
lealgamb: nit: `<String>` to be consistent with code from other classes.

src/test/java/codeu/model/data/ConversationTest.java
```diff
@@ -27,12 +28,32 @@ public void testCreate() {
...
 
     Assert.assertEquals(id, conversation.getId());
     Assert.assertEquals(owner, conversation.getOwnerId());
     Assert.assertEquals(title, conversation.getTitle());
     Assert.assertEquals(creation, conversation.getCreationTime());
+    Assert.assertEquals(isPrivate, conversation.isPrivate());
+    Assert.assertEquals(members, conversation.getMembers());
+    Assert.assertEquals(containsTestUser1, conversation.containsMember("testUser1"));
+
+    conversation.removeMember("testUser1");
```
lealgamb: nice thinking on mocking the add/remove

jasminechau2:
This is awesome

src/main/java/codeu/model/store/persistence/PersistentDataStore.java
vocatwk: No unfortunately.

src/main/webapp/WEB-INF/view/addUserBox.jsp
vocatwk: Good idea.

src/main/java/codeu/model/store/persistence/PersistentDataStore.java
```diff
@@ -107,9 +108,11 @@ public PersistentDataStore() {
...
 
         Conversation conversation = new Conversation(uuid, ownerUuid, title, creationTime);
         if(isPrivate != null && isPrivate == true) conversation.makePrivate();
+        if(membersAsList != null) conversation.addMembers(new HashSet<>(membersAsList));
```
yebrahim: nit: (here and the previous line) I'd prefer breaking the if condition's body to next line, preferably with braces. It's much more readable that way, and avoids incorrect edits in the future.

src/main/webapp/WEB-INF/view/chat.jsp
```diff
@@ -39,18 +39,41 @@ String privacySettingButtonValue = (Boolean) request.getAttribute("isPrivate")?
     }
   </style>
 
+  <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.3.1/jquery.min.js"> </script>
```
yebrahim: Now that we're relying on jquery, how about we load it only once, perhaps in the index page? Once it's load there, it should be accessible everywhere.

src/main/webapp/WEB-INF/view/chat.jsp
```diff
@@ -39,18 +39,41 @@ String privacySettingButtonValue = (Boolean) request.getAttribute("isPrivate")?
...
+
   <script>
+
+    //open add user box
+    $(document).ready(function(){
+      $("#adder").click(function() {
```
yebrahim: Notice the `#addr` selector is evaluated on the entire document. If another element has the same name, which is general enough, it'll cause trouble. Let's make these names more specific? Maybe "addUsersModalTrigger" or something.

src/main/webapp/WEB-INF/view/chat.jsp
```diff
@@ -39,18 +39,41 @@ String privacySettingButtonValue = (Boolean) request.getAttribute("isPrivate")?
...
+    $(document).ready(function() {
+      $(".close").click(function() {
+        $("#addUserBox").hide();
+      });
+    });
+
+    // show contents of drop down settings menu
+    function toggleSettingsDropdown(){
+      var dropDiv = document.getElementsByClassName("dropdown-content")[0];
+      if(dropDiv.style.display === "none"){
```
yebrahim: If you're using jquery, I think there's a `.toggle` function that can replace these five lines.

src/main/webapp/WEB-INF/view/chat.jsp
```diff
@@ -60,17 +83,27 @@ String privacySettingButtonValue = (Boolean) request.getAttribute("isPrivate")?
...
     <h1>
+      <!-- Conversation title -->
       <%= conversation.getTitle() %>
-      <i onclick="toggleSettingsDropdown()" class="fa fa-cog"> </i>
+        
+      <!-- Setting button and content -->
+      <div id="dropdown-settings">
+        <i onclick="toggleSettingsDropdown()" class="fa fa-cog"> </i>
+        <div class="dropdown-content" style="display:none">
+          <li> <form action="/chat/<%= conversation.getTitle() %>" method="POST">
```
yebrahim: Now that we're doing AJAX calls (using the `fetch` API), how about we replace these forms with simple buttons that call the endpoint using `fetch`?

src/main/webapp/WEB-INF/view/chat.jsp
vocatwk: The drop downs I access using jquery aren't working when I move that line to the index page. I'm not sure what i'm doing wrong.

src/main/webapp/WEB-INF/view/chat.jsp
vocatwk: Oops yeah I forgot to change that.

src/main/webapp/WEB-INF/view/chat.jsp
```diff
@@ -58,15 +57,34 @@ String privacySettingButtonValue = (Boolean) request.getAttribute("isPrivate")?
...
+    });
+
+    // for make private/make public button
+    var newChatPrivacyValue = "<%= privacySettingButtonValue %>";
+    $(document).ready(function() {
+      $("#privacySettingButton").click(function() {
+        fetch('/chat/<%= conversation.getTitle() %>', {
+          method: "POST",
+          headers: {
+            "Purpose": newChatPrivacyValue,
```
yebrahim: What you want to use here is the `body` field of the POST request, instead of the headers. This is what is canonically used to pass the contents of the request, and it can be an arbitrary object (serialized using JSON). See: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#Body.

PR #78, opened 2018-07-03
Bot now responds when it's @ mentioned.

lealgamb:
I created a class for the `Bot`. The class has the basic functions like parsing through the user's message and generating a response. Currently the Bot only returns a generic answer, so I'll be working on adding the keywords. 

Creating a `Message` with a UUID was tricky, and so was getting the Bot's "username" so I'm using the UUID `0000000-0000-0000-0000-000000000000` for now. Apparently the probability of UUID's colliding is low according to some searching I did, but there's probably a better way to do this. 

I'm also planning on creating an HTML page for the Bot to make it seem like a real user and to be able to access its profile page. 

I'll add tests in a future PR.

image

The Activity Feed was also updated accordingly:
image


src/main/java/codeu/controller/ChatServlet.java
```diff
@@ -170,12 +171,34 @@ else if(messageContent != null) {
...
-      List<String> messageInformation = new ArrayList<>();
+      List<String> messageInformation = new ArrayList<String>();
       messageInformation.add(user.getName());
       messageInformation.add(conversationTitle);
       messageInformation.add(cleanedMessageContent);
       Event messageEvent = new Event(UUID.randomUUID(), "Message", message.getCreationTime(), messageInformation);
       eventStore.addEvent(messageEvent);
+
+      // Scan the message for "@NemoBot"
+      if (cleanedMessageContent.toLowerCase().contains("@NemoBot".toLowerCase())) {
```
yebrahim: We don't just want this to be contained in the message, we need it to be a standalone word right? Something like "@NemoBotzzz" shouldn't trigger the bot.

src/main/java/codeu/controller/ChatServlet.java
```diff
@@ -170,12 +171,34 @@ else if(messageContent != null) {
...
+                botResponse,
+                Instant.now());
+
+        messageStore.addMessage(botMessage);
+
+        List<String> botMessageInformation = new ArrayList<String>();
+        botMessageInformation.add("NemoBot");
+        botMessageInformation.add(conversationTitle);
+        botMessageInformation.add(botResponse);
+        Event botMessageEvent = new Event(UUID.randomUUID(), "Message", botMessage.getCreationTime(), botMessageInformation);
```
yebrahim: nit: can we break up long lines to help with readability?

src/main/java/codeu/model/data/Bot.java
```diff
@@ -0,0 +1,43 @@
+package codeu.model.data;
```
yebrahim: Since we're planning to have a profile page for the bot, how about making this extend the `User` class? It's reasonable to think of this as a specialized type of user, and we get the `id`, `name`, `creation`, and `aboutMe` (and maybe even `isAdmin`?) fields for free, which all make sense to me to have on a `Bot` instance. This way we also don't have to handle it differently on the chat JSP.

src/test/java/codeu/controller/SearchServletTest.java
```diff
@@ -0,0 +1,59 @@
+package codeu.controller;
```
yebrahim: A trick to keep the PR purely for the bot changes, when the search servlet tests haven't been merged yet is to change the base branch of the PR, from master to the branch of the search tests. This means you're now building on top of that branch instead of master. Note that this means that if you merge the base branch first into master, you'll need to change this PR's base branch back to master to continue diffing against the right set of changes, and to actually merge into master when done.

If that's too confusing it's fine, you don't have to do it. Github doesn't unfortunately handle multiple PRs building on top of each other very well. :)

src/main/java/codeu/controller/ChatServlet.java
lealgamb: Right. Would requiring the user to @ mention the bot as the first word in the message be a good way to go about this? That way it'll be simpler to check the message until the first white space?

src/main/java/codeu/model/data/Bot.java
lealgamb: I hadn't thought of this, thanks! For this approach, would it still make sense to assign the same `id` on each `Bot` instance?

src/test/java/codeu/controller/SearchServletTest.java
lealgamb: Thanks for this tip!

src/main/java/codeu/controller/ChatServlet.java
lealgamb: Will do, sorry about that.

lealgamb:
Quick note: I added the favicon in this PR

src/main/java/codeu/controller/ChatServlet.java
yebrahim: I'd say that's a reasonable initial restriction. One we might decide to relax later on, but good enough for now.

src/main/java/codeu/model/data/Bot.java
yebrahim: Not really, bots are different entities, much like different users, so they'd get a unique id each in this case.

yebrahim:
I'd like to discuss the extensibility of the framework being developed here. How do you think we can design this to allow developers to easily add new bot implementations?
@lealgamb @jasminechau2 @vocatwk to pitch in. (Feel free to move this discussion to a separate thread if you find it too much for this PR).

src/main/java/codeu/controller/ChatServlet.java
```diff
@@ -145,7 +146,9 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
     }
 
     String purpose = request.getReader().readLine();
+    System.out.println("purpose: " + purpose);
```
vocatwk: Don't forget to remove this + line 151

yebrahim:
Nice touch with the favicon! :)

lealgamb:
The @ mention is now restricted to a standalone word.

src/main/java/codeu/controller/ChatServlet.java
```diff
@@ -159,12 +162,45 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+          new Event(
+              UUID.randomUUID(), 
+              "Message", 
+              message.getCreationTime(), 
+              messageInformation);
       eventStore.addEvent(messageEvent);
+
+      // Scan the message for "@NemoBot"
+      if (containsExactMatch(cleanedMessageContent, "@NemoBot")) {
+        System.out.println("Contains exact match");
```
yebrahim: Remove?

src/main/java/codeu/controller/ChatServlet.java
```diff
@@ -159,12 +162,45 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+      Event messageEvent = 
+          new Event(
+              UUID.randomUUID(), 
+              "Message", 
+              message.getCreationTime(), 
+              messageInformation);
       eventStore.addEvent(messageEvent);
+
+      // Scan the message for "@NemoBot"
+      if (containsExactMatch(cleanedMessageContent, "@NemoBot")) {
```
yebrahim: nit: How about we call this `containsWholeWord` since it's not exact match; it ignores case.

src/main/java/codeu/controller/ChatServlet.java
```diff
@@ -197,4 +233,17 @@ else if(purpose.equals("make public")){
     }
     conversationStore.updateConversation(conversation);
   }
+
+  /*
+  * This function splits source by whitespace and checks if substring 
+  * is contained as a standalone word, ignore case.
+  */
+  private boolean containsExactMatch(String source, String substring) {
+    for (String word : source.split("\\s+")) {
```
yebrahim: Note punctuation won't work with this implementation. Something like `@bot: something` or `hi @bot!` won't work.
Maybe we should stick with assuming the mention will be the first word? This is simpler, and it's also a reasonable restriction. If you still want this flexibility, you'll need to also split on punctuation characters. See this answer: https://stackoverflow.com/a/7385150/1768702.

lealgamb:
Here's an update on what the bot can respond to:
image

PR #86, opened 2018-07-08
Notifications Test

jasminechau2:
This is the test for the notification servlet


src/test/java/codeu/controller/NotificationServletTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    UUID fakeConversationId = UUID.randomUUID();
+    Conversation fakeConversation = new Conversation(fakeConversationId, fakeUserId, "testConversation", Instant.ofEpochMilli(1000));
+    List<String> conversationInformation = new ArrayList<>();
+    conversationInformation.add(fakeUser.getName());
+    conversationInformation.add(fakeConversation.getTitle());
+    Event conversationEvent = new Event(UUID.randomUUID(), "Conversation", fakeConversation.getCreationTime(), conversationInformation);
+    fakeEventList.add(conversationEvent);
+
+    Message fakeMessage = new Message(UUID.randomUUID(), fakeConversationId, fakeUserId, "testMessage", Instant.ofEpochMilli(1000));
```
yebrahim: Do all events have the same creation time: 1000?

src/test/java/codeu/controller/NotificationServletTest.java
```diff
@@ -0,0 +1,108 @@
...
+    List<String> messageInformation = new ArrayList<>();
+    messageInformation.add(fakeUser.getName());
+    messageInformation.add(fakeConversation.getTitle());
+    messageInformation.add(fakeMessage.getContent());
+    Event messageEvent = new Event(UUID.randomUUID(), "Message", fakeMessage.getCreationTime(), messageInformation);
+    fakeEventList.add(messageEvent);
+
+    Instant fakeLastSeenTime = Instant.now();
+    mockUser.setLastSeenNotifications(fakeLastSeenTime);
```
yebrahim: How about we make the fake events have different times (e.g. 1000, 2000, and 3000), then set the user's last seen time to 1500 or something, then we should expect that the last two events only are returned? This way we're actually testing that the user only sees the new events.

src/test/java/codeu/controller/NotificationServletTest.java
```diff
@@ -0,0 +1,108 @@
...
+    Event messageEvent = new Event(UUID.randomUUID(), "Message", fakeMessage.getCreationTime(), messageInformation);
+    fakeEventList.add(messageEvent);
+
+    Instant fakeLastSeenTime =  Instant.ofEpochMilli(1500);
+    mockUser.setLastSeenNotifications(fakeLastSeenTime);
+
+
+    Mockito.when(mockSession.getAttribute("user")).thenReturn("test_User");
+    Mockito.when(mockUserStore.getUser("test_User")).thenReturn(mockUser);
+    Mockito.when(mockEventStore.getEventsSince(fakeLastSeenTime)).thenReturn(fakeEventList);
```
yebrahim: If we don't mock the call to `getEventsSince`, then we should expect the `doGet` call to return just the last two events, since they're both past 1500 ms, right?

src/test/java/codeu/controller/NotificationServletTest.java
```diff
@@ -0,0 +1,112 @@
...
+    List<String> messageInformation = new ArrayList<>();
+    messageInformation.add(fakeUser.getName());
+    messageInformation.add(fakeConversation.getTitle());
+    messageInformation.add(fakeMessage.getContent());
+    Event messageEvent = new Event(UUID.randomUUID(), "Message", fakeMessage.getCreationTime(), messageInformation);
+    fakeEventList.add(messageEvent);
+    eventsToShow.add(messageEvent);
+
+    Instant fakeLastSeenTime = Instant.ofEpochMilli(100);
+    mockUser.setLastSeenNotifications(fakeLastSeenTime);
```
vocatwk: Not sure why this is needed if you're mocking the call to the getter function on line 98.

src/test/java/codeu/controller/NotificationServletTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+    Mockito.when(mockEventStore.getAllEvents()).thenReturn(fakeEventList);
+    Mockito.when(mockSession.getAttribute("user")).thenReturn("fakeUser");
+    Mockito.when(mockUserStore.getUser("fakeUser")).thenReturn(mockUser);
+    Mockito.when(mockUser.getLastSeenNotifications()).thenReturn(fakeLastSeenTime);
+
+    Mockito.when(mockEventStore.getEventsSince(fakeLastSeenTime)).thenReturn(eventsToShow);
+    Mockito.verify(mockUser).setLastSeenNotifications(fakeLastSeenTime);
```
vocatwk: Not sure why you're verifying this before the doGet call. Also, should it be set with fakeLastSeenTime?

src/test/java/codeu/controller/NotificationServletTest.java
```diff
@@ -0,0 +1,116 @@
...
+    List<String> messageInformation = new ArrayList<>();
+    messageInformation.add(fakeUser.getName());
+    messageInformation.add(fakeConversation.getTitle());
+    messageInformation.add(fakeMessage.getContent());
+    Event messageEvent = new Event(UUID.randomUUID(), "Message", fakeMessage.getCreationTime(), messageInformation);
+    fakeEventList.add(messageEvent);
+    eventsToShow.add(messageEvent);
+
+
+    Instant fakeLastSeenTime = Instant.ofEpochMilli(4000);
```
lealgamb: Wouldn't using 4000 result in 0 notifications?

src/test/java/codeu/controller/NotificationServletTest.java
```diff
@@ -0,0 +1,111 @@
...
+    eventsToShow.add(messageEvent);
+
+
+    Instant fakeLastSeenTime = Instant.ofEpochMilli(1500);
+
+    Mockito.when(mockSession.getAttribute("user")).thenReturn("fakeUser");
+    Mockito.when(mockUserStore.getUser("fakeUser")).thenReturn(mockUser);
+    Mockito.when(mockEventStore.getAllEvents()).thenReturn(fakeEventList);
+    Mockito.when(mockUser.getLastSeenNotifications()).thenReturn(fakeLastSeenTime);
+    Mockito.when(mockEventStore.getEventsSince(fakeLastSeenTime)).thenReturn(eventsToShow);
```
yebrahim: Why do we mock this? If we're manually inserting three events at 1000, 2000, and 3000 milliseconds, then this call should automatically work when the `doGet` method calls it. This test should be making sure the verify below on line 105 works when this isn't mocked.

src/main/java/codeu/controller/NotificationServlet.java
```diff
@@ -46,10 +46,18 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
         String username = (String)request.getSession().getAttribute("user");
         User user = userStore.getUser(username);
         Instant lastSeenTime = user.getLastSeenNotifications();
-        Instant userLookedAtPage = Instant.now();
 
         List<Event> eventsToShow = eventStore.getEventsSince(lastSeenTime);
-        user.setLastSeenNotifications(userLookedAtPage);
+        int lastEventPlace = 1;
+        
+        if(eventsToShow.size() > lastEventPlace){
```
yebrahim: I'm not sure what this is doing. Why did we change how we store the last seen time?

jasminechau2: I was having a hard time with testing due to making the userLookedAtPage = Instant.now() because the time never matched up exactly. This takes the last event the user sees and takes it's creation time making it the last seen time. The if statement checks if the eventsToShow is empty or not and if the size is bigger then one it takes the last element of the list and sets the notification time as creation.

jasminechau2: I can add a comment on the servlet

yebrahim: Can we mock `Instant.now`? Maybe something like https://stackoverflow.com/a/44795939/1768702.

jasminechau2: I can try that

src/test/java/codeu/controller/NotificationServletTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+    Instant fakeLastSeenTime = Instant.ofEpochMilli(1500);
+
+    Mockito.when(mockSession.getAttribute("user")).thenReturn("fakeUser");
+    Mockito.when(mockUserStore.getUser("fakeUser")).thenReturn(mockUser);
+    Mockito.when(mockEventStore.getAllEvents()).thenReturn(fakeEventList);
+    Mockito.when(mockUser.getLastSeenNotifications()).thenReturn(fakeLastSeenTime);
+
+    Mockito.when(mockEventStore.getEventsSince(fakeLastSeenTime)).thenReturn(spyEvents);
+    Mockito.doReturn(2).when(spyEvents).size();
```
yebrahim: Do we really need to mock this? The list does have two items, shouldn't calling its size already return 2?

src/test/java/codeu/controller/NotificationServletTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+    Instant currentTime = Instant.now();
+    long nanosTosubtract = currentTime.getNano();
+    Instant lastEventTime = currentTime.minusNanos(nanosTosubtract);
+
+    Mockito.when(mockClock.instant()).thenReturn(lastEventTime);
+
+    NotificationServlet.doGet(mockRequest, mockResponse);
+    Assert.assertEquals(2, spyEvents.size());
```
yebrahim: This doesn't really add anything, can we just delete it?

src/main/java/codeu/controller/NotificationServlet.java
```diff
@@ -43,13 +46,16 @@ void setEventStore(EventStore eventStore){
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response)
       throws IOException, ServletException {
+        Instant CurrentTime = clock.instant();
```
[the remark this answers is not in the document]
jasminechau2: Having the nanoseconds makes the test fail because servlet and the testing current time never matches up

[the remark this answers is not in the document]
yebrahim: Shouldn't the mock take care of that?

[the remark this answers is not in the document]
jasminechau2: I tried using the stackoverflow example but still didnt to get the same time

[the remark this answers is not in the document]
yebrahim: What error do you get if you just use `CurrentTime` here, and `currentTime` in the servlet test file?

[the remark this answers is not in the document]
jasminechau2: screen shot 2018-07-17 at 12 20 07 am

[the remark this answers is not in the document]
yebrahim: I see. I'm fine with comparing without nano seconds for now then.

PR #87, opened 2018-07-08
Handled empty and null conversation names

vocatwk:
I added tests for the above two cases as well. 
screen shot 2018-07-08 at 2 40 10 am


vocatwk:
Thanks. Yeah that seems more organized.

PR #89, opened 2018-07-09
Bot Framework

### LINKED ISSUE #88: Set Up Bot Framework
- [x] Make a `Bot` interface

- [x] Rework the current `Bot` class

- [x] Make a `BotController` class

- [x] Update `ChatServlet` accordingly

lealgamb:
I made a `Bot` interface, which is very simple. Perhaps more stuff can be added later on. The former `Bot` class was renamed to `NemoBot`. Most of the new work is in `BotController`. I wanted to send this PR to get feedback before updating `ChatServlet`. The same goes for tests - I'll add them in a future PR once the initial framework changes are complete. 

EDIT: I semi-updated `ChatServlet` to work around the name changes. The changes don't reflect the addition of `BotController`.

Resolves #88

### DIFF
--- a/src/main/java/codeu/controller/BotController.java
+++ b/src/main/java/codeu/controller/BotController.java
@@ -39,9 +39,6 @@ public class BotController {
   * @return null if the given String does not match any existing Bot. 
   */
   public Bot getBot(String mentionKey) {
-    if (botMap.containsKey(mentionKey)) {
-      return botMap.get(mentionKey);
-    }
-    return null;
+    return botMap.get(mentionKey);
   }
 }
--- a/src/main/java/codeu/controller/BotController.java
+++ b/src/main/java/codeu/controller/BotController.java
@@ -13,7 +13,7 @@ public class BotController {
   /** Singleton instance of BotController. */
   private static BotController instance;
 
-  /** Map that stores a mention key to it's Bot. */
+  /** Map that stores a mention key to its Bot. */
   private Map<String, Bot> botMap;
 
   private BotController() {
--- a/src/main/java/codeu/controller/BotController.java
+++ b/src/main/java/codeu/controller/BotController.java
@@ -1,0 +1,47 @@
+package codeu.controller;
+
+import codeu.model.data.Bot;
+import java.util.Map;
+import java.util.HashMap;
+
+/**
+ * Controller class is responsible for controlling all Bots. 
+ * It's a singleton so all servlet classes can access the same instance.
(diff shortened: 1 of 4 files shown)


src/main/java/codeu/controller/BotController.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Controller class is responsible for controlling all Bots. 
+ * It's a singleton so all servlet classes can access the same instance.
+ */
+public class BotController {
+
+  /** Singleton instance of BotController. */
+  private static BotController instance;
+
+  /** Map that stores a mention key to it's Bot. */
```
yebrahim: nit: "its"

src/main/java/codeu/model/data/Bot.java
```diff
@@ -1,49 +1,9 @@
 package codeu.model.data;
 
-import codeu.model.store.basic.UserStore;
-import java.time.Instant;
-import java.util.UUID;
-import org.mindrot.jbcrypt.BCrypt;
+/** Interface for Bot implementations. */
+public interface Bot {
```
yebrahim: Thinking a little more about it, how do you think about making the Bot interface contain a `mentionKeyword` field that each Bot implementation populates, then the `registerBot` can just be called with the Bot instance?
The reason I'm suggesting this is to also make the Bot aware of its own mention keyword, which seems reasonable to me. You can imagine the About Me section of the Bot for example listing its mention keyword.
It also kind of cruel to make Bots that don't know their own name. :)

lealgamb: Haha, yes I'll add that, thanks for pointing this out. Does this also mean that using a set instead of a map would be more appropriate?

lealgamb: Also, if we include that field I'm guessing an abstract class would be used over an interface, right?

yebrahim: We'd still need a map because the registry needs to look up bots by the mention key. Sets are only useful to check if a certain bot is registered or not, i.e. look up by instance.
You can still use an interface that contain a string field. I know you can do that in other languages at least, but I suspect Java supports that too.

yebrahim: Ah, that seems like a limitation in Java. How about a `getName` method instead then?

lealgamb: Added 👍

PR #91, opened 2018-07-11
Implemented add button and Add users button in Add users modal

vocatwk:
Clicking the add button next to a user's name will now select a user to be added to the group. Clicking add users will send a PUT request with the list of users selected, close the modal and clear the list. This is just getting the list of selected user's to the chat servlet, I haven't done anything with that list yet.

I'm working on adding a close icon on each tag that shows up on top of the search bar so users can deselect and having the add users button deactivated atleast until one user is selected.

I also added a header to differentiaite between changing chat privacy PUT request and adding users PUT request.

screen shot 2018-07-11 at 3 36 03 am


src/main/java/codeu/controller/ChatServlet.java
```diff
@@ -220,15 +220,28 @@ public void doPut(HttpServletRequest request, HttpServletResponse response)
       return;
     }
 
-    String purpose = request.getReader().readLine();
+    String purpose = request.getHeader("purpose");
+    if(purpose == null){
```
lealgamb: Nice job using the header to determine the needed action, looks clean and it's easy to understand. 👍

src/main/webapp/css/main.css
```diff
@@ -124,4 +124,16 @@ button {
...
+#selectedUserList {
+  display: flex;
+}
+
+.selectedUser {
+  width: 50px;
+  height: 25px;
+  padding: 0;
+  font-size: 12px;
+  align-content:;
```
yebrahim: Remove

src/main/webapp/css/main.css
```diff
@@ -124,4 +124,16 @@ button {
   display: flex;
 }
 
+#selectedUserList {
+  display: flex;
+}
+
+.selectedUser {
+  width: 50px;
```
yebrahim: Try adding a `margin-right: 5px;` line here.

src/main/webapp/WEB-INF/view/searchbar.jsp
```diff
@@ -24,6 +24,9 @@
                   a2 = document.createElement("a");
                   a2.style = "float: right";
                   a2.innerHTML = "add";
+                  a2.href = "#";
+                  a2.setAttribute("class", "add-user-button");
```
yebrahim: What if the user is already added? This means we can add users more than once if we search, add, clear, then repeat.

src/main/webapp/WEB-INF/view/chat.jsp
```diff
@@ -63,6 +66,53 @@ String privacySettingButtonValue = (Boolean) request.getAttribute("isPrivate")?
...
+          headers: {
+            "Content-Type": "application/json; charset=utf-8",
+            "purpose" : "Adding users"
+          },
+          body: JSON.stringify(selectedUserNames),
+          credentials: "same-origin"
+        }).then(function(response) {
+          if(!response.ok) {
+            console.log("An error occured. Status code: " + response.status);
+            return;
```
yebrahim: You don't need the return statement here.

src/main/webapp/WEB-INF/view/searchbar.jsp
```diff
@@ -24,6 +24,9 @@
...
+                  a2.setAttribute("id", userName);
```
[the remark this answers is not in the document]
vocatwk: Oh I didn't know we could do that. Thanks.

src/main/webapp/WEB-INF/view/searchbar.jsp
vocatwk: I was thinking of taking care of that on the server side(not adding it if the user is already there) but I could disable that button for users already in that conversation as well.

yebrahim: The server does not have this context though, right? The client code is compiling the list of users to be added to the conversation, so only it knows about who's in that list, before it sends it off to the server.

vocatwk: Wouldn't the server have access to the list of members in that conversation? I could disable the button for users already in that conversation just like it's disabled now if it's already selected then when it sends that list of names, I could add it to the lists of members.

yebrahim: Oh, what I meant was the list of currently selected-for-addition users. Where you click the "Add" button and it queues it up? If you add a few users, clear the search, then search again, the "Add" buttons are all enabled for the new list, so you can add them again.

vocatwk: Oh, sorry I misunderstood that. I'll find a way to fix that.

vocatwk:
Okay, I'll try and find a better way.

src/main/webapp/WEB-INF/view/searchbar.jsp
yebrahim: It's not a simple problem given the current design of the search box, so feel free to discuss your ideas here. :)

vocatwk:
I was thinking of disabling the add Users button when the search results are populated if the user is already a member or if it's already queued up. I could maintain a list of the ones that are queued up whenever the modal opens and find a way to clear this list when the modal closes. What do you guys think?

yebrahim:
If you can maintain a list of the queued up users, so that you can check before adding the "Add" button when the search results come back, that would be a good fix.
Where were you thinking of maintaining such a list?

vocatwk:
Well I haven't tried it out yet(I'm about to) but I was thinking of initializing an empty list when the page first loads in chat.jsp and then adding users to it accordingly.

vocatwk:
Another option would be to set up a servlet for it or use one of the existing ones and use the fetch api to add to a static list and fetch that list right before the add button is added to check if it should be enabled or not.

src/main/webapp/WEB-INF/view/chat.jsp
vocatwk: I didn't remove this because I wanted to empty the list upon adding users to the list.

PR #93, opened 2018-07-11
Updated ChatServlet based on BotController

### LINKED ISSUE #105: Bot Messages Don't Show on Activity Feed
Realized that bot messages don't show up on the activity feed. They do show up in notifications so that means they're registering as events. 

image
image

lealgamb:
`ChatServlet` now reflects the addition of `BotController`. I added some methods to the `Bot` interface which `NemoBot` inherits from `User`. `SeverStartpListener` is where bots are registered. 

`ChatServletTest` was updated as well. However, I couldn't fully test `getMentionKey()` because it's a private function. I can probably just copy the code and "hard test" it. Let me know what you guys think about this. 

One last thing, I updated the search for users code to ignore case. Now, users can also search for bots since they're technically also users.

References #88 

Fixes #105


lealgamb:
Merged the password length check here instead of master, sorry.

src/main/java/codeu/controller/ServerStartupListener.java
```diff
@@ -36,6 +45,10 @@ public void contextInitialized(ServletContextEvent sce) {
       List<Event> events = PersistentStorageAgent.getInstance().loadEvents();
       EventStore.getInstance().setEvents(events);
 
+      NemoBot nemoBot = new NemoBot();
+      botController.registerBot("@NemoBot", nemoBot);
```
yebrahim: We can use the bot name here instead of hardcoding it twice:

`botController.registerBot(nemoBot.getName(), nemoBot);`

We can add the `@` when looking up bots with the mention key.

yebrahim: Can `registerBot` take care of adding it as a user too?

src/test/java/codeu/controller/BotControllerTest.java
```diff
@@ -0,0 +1,56 @@
...
+  public void setup() {
+    botController = BotController.getTestInstance();
+
+    Map<String, Bot> botMap = new HashMap<String, Bot>();
+    botMap.put("@NEMO_BOT", NEMO_BOT);
+    botController.setBots(botMap);
+  }
+
+  @Test
+  public void testGetBot_found() {
```
yebrahim: Add a test where mention key in lower case?

src/main/java/codeu/controller/ServerStartupListener.java
lealgamb: I tried this and it doesn't work. I get the error `incompatible types: codeu.model.data.Bot cannot be converted to codeu.model.data.User`.

lealgamb: Never mind, casting `Bot` as `User` worked.

lealgamb: However, the tests now say `No API environment is registered for this thread.`

yebrahim: That seems like a weird error for such change. Have you googled around?

yebrahim: Seems like you've fixed this in the other PR? I can see the bot is registering itself as a user there.

PR #99, opened 2018-07-15
Added queued up users to conversation

vocatwk:
I also update the list of members each time. The add button is now only on users that aren't members of the conversation already.

### DIFF
--- a/src/main/java/codeu/controller/ChatServlet.java
+++ b/src/main/java/codeu/controller/ChatServlet.java
@@ -29,12 +29,15 @@ import java.time.Instant;
 import java.util.List;
 import java.util.ArrayList;
 import java.util.UUID;
+import java.util.HashSet;
+import java.util.Arrays;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import org.jsoup.Jsoup;
 import org.jsoup.safety.Whitelist;
+import com.google.gson.Gson;
 
 /** Servlet class responsible for the chat page. */
 public class ChatServlet extends HttpServlet {
@@ -238,7 +241,15 @@ public class ChatServlet extends HttpServlet {
       conversationStore.updateConversation(conversation);
     }
     else if(purpose.equals("Adding users")){
-      //TODO parse json here
+
+      String[] userNameArray = stringToArray(request.getReader().readLine());
+      HashSet<String> toBeAdded = new HashSet<>(Arrays.asList(userNameArray));
+      conversation.addMembers(toBeAdded);
+
+      String json = new Gson().toJson(conversation.getMembers());
+      response.setContentType("application/json");
+      response.setCharacterEncoding("UTF-8");
+      response.getWriter().write(json);
     }
     
   }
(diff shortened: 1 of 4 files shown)


src/main/webapp/WEB-INF/view/chat.jsp
```diff
@@ -75,20 +75,53 @@ String membersOfConversation = (String) request.getAttribute("membersOfConversat
...
         selectedUser.setAttribute("class", "btn btn-primary");
         selectedUser.classList.add('selectedUser');
-        selectedUser.innerHTML = this.getAttribute("username");
+        selectedUser.innerHTML = userName;
+
+        var closeIcon = document.createElement("button");
+        closeIcon.setAttribute("type", "button");
+        closeIcon.setAttribute("class", "close");
+        closeIcon.setAttribute("aria-label", "Close");
+        closeIcon.setAttribute("id", "remove-from-queue-button");
```
yebrahim: We shouldn't be setting the ID of more than one element to the same value. A couple of alternative solutions here: Either add a click listener to this button before adding it:
`closeIcon.addEventListener('click', () => $(selectedUser).remove();`
Or use the class to listen on element clicks, like you did on line 74, and add a custom attribute `username`, also like you did for the add user button.
The first solution is both easier and cleaner.

vocatwk: Thanks, will fix it now.

src/main/java/codeu/controller/ChatServlet.java
```diff
@@ -238,7 +244,16 @@ else if(privacyCommand.equals("make public")){
       conversationStore.updateConversation(conversation);
     }
     else if(purpose.equals("Adding users")){
-      //TODO parse json here
+
+      String[] userNameArray = stringToArray(request.getReader().readLine());
```
yebrahim: You shouldn't need to parse JSON yourself, Gson should do that for you. Something like: https://stackoverflow.com/a/18421994/1768702.

yebrahim: Whether you'll take my suggestion above about using Gson for parsing, or you want to do parsing yourself, you should be very careful about handling input coming from the client, and at least add `try/catch` logic around it. Remember clients can send anything to the server, including malformed JSON, and the server should never crash on bad input, which is what'll happen here probably.

We should also sanitize usernames before adding them by both: cleaning up any HTML in them (to avoid SQL injection threats and the like), and making sure they are valid user names of actual users in the system, so we don't clutter the conversation's member list with incorrect names that might break later on NPEs when fetching.

Let me know if this doesn't make sense or if we should discuss more.

src/main/webapp/WEB-INF/view/searchbar.jsp
```diff
@@ -21,20 +21,23 @@
...
-                  a2 = document.createElement("a");
-                  a2.style = "float: right";
-                  a2.innerHTML = "add";
-                  a2.setAttribute("class", "add-user-button");
-                  a2.setAttribute("username", userName);
+                  li.appendChild(a1);
+
+                  if(!membersOfConversation.has(userName)){
+                    a2 = document.createElement("a");
+                    a2.style = "float: right";
```
yebrahim: This can go into the `add-user-button` class.

src/main/java/codeu/controller/ChatServlet.java
vocatwk: Oh I thought you had to have a specific class for to parse into. Didn't know you could it to an array. Thanks!

src/main/java/codeu/controller/ChatServlet.java
```diff
@@ -245,7 +245,25 @@ else if(privacyCommand.equals("make public")){
...
-      String[] userNameArray = stringToArray(request.getReader().readLine());
+      String jsonString = request.getReader().readLine();
+      String cleanedJsonString = Jsoup.clean(jsonString, Whitelist.none());
+      String[] userNameArray = null;
+      
+      try{
+        userNameArray = new Gson().fromJson(cleanedJsonString, String[].class);
+      }
+      catch(Exception e){
+        response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
```
yebrahim: Awesome!

PR #100, opened 2018-07-16
Allow users to create conversations with multi-word titles

### LINKED ISSUE #85: Can't create conversations with multiple words (space-separated) as title
This strikes me as a weird restriction, do you guys agree we should relax it?

vocatwk:
resolves #85 

I encode spaces in the conversation title with "-" when stored in the conversation class. I replace them back with a space during display. I was going to do it the other way(replace the spaces with "-" when sending requests and putting them back after) but I thought this way would be easier and less riskier.

Also, I noticed we don't clean conversation titles for html like we do messages, is there a reason for that or should we add it?

EDIT

I switched to using UUIDs for the urls, I updated the tests and the Event information as well as removed some restrictions.

### DIFF
--- a/src/main/java/codeu/controller/ChatServlet.java
+++ b/src/main/java/codeu/controller/ChatServlet.java
@@ -107,16 +107,7 @@ public class ChatServlet extends HttpServlet {
 
     String requestUrl = request.getRequestURI();
     String conversationIdAsString = requestUrl.substring("/chat/".length());
-    UUID conversationId = null;
-
-    try{
-      conversationId = UUID.fromString(conversationIdAsString);
-    }
-    catch(Exception e){
-      System.out.println("Conversation was null: " + conversationIdAsString);
-      response.sendRedirect("/conversations");
-      return;
-    }
+    UUID conversationId = getIdFromString(conversationIdAsString);
 
     Conversation conversation = conversationStore.getConversation(conversationId);
     if (conversation == null) {
@@ -152,16 +143,7 @@ public class ChatServlet extends HttpServlet {
 
     String requestUrl = request.getRequestURI();
     String conversationIdAsString = requestUrl.substring("/chat/".length());
-    UUID conversationId = null;
-
-    try{
-      conversationId = UUID.fromString(conversationIdAsString);
-    }
-    catch(Exception e){
-      System.out.println("Conversation was null: " + conversationIdAsString);
-      response.sendRedirect("/conversations");
-      return;
-    }
+    UUID conversationId = getIdFromString(conversationIdAsString);
 
     Conversation conversation = conversationStore.getConversation(conversationId);
     if (conversation == null) {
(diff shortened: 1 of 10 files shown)


vocatwk:
Oh that does sound like a much better idea, I'll change it to using UUID

lealgamb:
I like the idea of using UUID. Also, I'm in favor of cleaning any html content.

src/main/java/codeu/controller/ChatServlet.java
```diff
@@ -104,19 +104,28 @@ void setEventStore(EventStore eventStore) {
...
   public void doGet(HttpServletRequest request, HttpServletResponse response)
       throws IOException, ServletException {
+
     String requestUrl = request.getRequestURI();
-    String conversationTitle = requestUrl.substring("/chat/".length());
+    String conversationIdAsString = requestUrl.substring("/chat/".length());
+    UUID conversationId = null;
 
-    Conversation conversation = conversationStore.getConversationWithTitle(conversationTitle);
+    try{
```
yebrahim: This seems to be duplicated below. Have a method to extract the id from the string?

src/main/java/codeu/controller/ConversationServlet.java
```diff
@@ -112,39 +112,25 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
 
-    if (!conversationTitle.matches("[\\w*\\s*]*")) {
-      request.setAttribute("error", "Please enter only letters and numbers.");
-      request.getRequestDispatcher("/WEB-INF/view/conversations.jsp").forward(request, response);
-      return;
-    }
-
-    String conversationTitleStored = conversationTitle.replaceAll(" ", "-");
-
-    if (conversationStore.isTitleTaken(conversationTitleStored)) {
```
yebrahim: Are we using this method anywhere else? If not, let's remove it from `ConversationStore`?

src/main/webapp/WEB-INF/view/activityfeed.jsp
```diff
@@ -47,18 +47,20 @@ List<Event> events = (List<Event>) request.getAttribute("events");
...
+                String conversationId = information.get(2);
               %>
                 <a href="/profile/<%= userName %>"><%= userName %></a> created a new conversation: 
-                <a href="/chat/<%= conversationTitle %>"><%= conversationTitle %></a>
+                <a href="/chat/<%= conversationId %>"><%= conversationTitle %></a>
           </li>
               <%
               }
-              else if (event.getType().equals("Message")) {
+              else if (event.getType().equals("Message") && information.size() > 3) {
```
yebrahim: Can we ever have the first part of the condition without the second? Are you thinking about backward compatibility (handling already existing data)?

src/test/java/codeu/controller/ConversationServletTest.java
```diff
@@ -84,49 +84,6 @@ public void testDoGet() throws IOException, ServletException {
     Mockito.verify(mockRequestDispatcher).forward(mockRequest, mockResponse);
   }
 
-  @Test
-  public void testDoPost_BadConversationName() throws IOException, ServletException {
```
yebrahim: Yay, removing tests!! 🎆 🎉

src/main/java/codeu/controller/ChatServlet.java
```diff
@@ -104,19 +104,28 @@ void setEventStore(EventStore eventStore) {
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response)
       throws IOException, ServletException {
+
     String requestUrl = request.getRequestURI();
-    String conversationTitle = requestUrl.substring("/chat/".length());
+    String conversationIdAsString = requestUrl.substring("/chat/".length());
+    UUID conversationId = null;
 
-    Conversation conversation = conversationStore.getConversationWithTitle(conversationTitle);
```
yebrahim: Remove this method if not used anywhere else?

src/main/webapp/WEB-INF/view/activityfeed.jsp
vocatwk: Yeah, I didn't want it to throw exceptions on the already existing events.

PR #101, opened 2018-07-17
Nemo Bot Can Reply

lealgamb:
NemoBot can reply to messages now. Its map is initialized based on `nemobot.txt`, so feel free to edit the file if you want. Also, the key of the map is in lower case, and so is the message parsing, to account for user spelling/punctuation. The same regex that scans for `@MentionKey` is used to account for punctuation. 

The main problem I see from this is what should happen when a user mentions multipled keywords. We can either have a restriction to only allow 1 keyword per message, or add n keywords in 1 message to a list and get all the replies into the bot's answer. What do you guys think?

I'll be working on tests for NemoBot soon, most likely in a different PR.

image


src/main/java/codeu/controller/ServerStartupListener.java
```diff
@@ -36,6 +42,9 @@ public void contextInitialized(ServletContextEvent sce) {
       List<Event> events = PersistentStorageAgent.getInstance().loadEvents();
       EventStore.getInstance().setEvents(events);
 
+      NemoBot nemoBot = new NemoBot("../../src/main/java/codeu/model/data/NemoBot.txt");
```
yebrahim: Will this path work when deployed?

src/main/java/codeu/model/data/NemoBot.txt
```diff
@@ -0,0 +1,17 @@
+Hi	Hey! I'm NemoBot. Want to talk to me? Simply @ mention me in any conversation!
```
yebrahim: Good work adding this file, makes for very clean separation of logic and data.

lealgamb:
Should I merge this to `BotBranch`, then merge PR #93 to `master`?

yebrahim:
Yes, if both PRs are ready for merging.

lealgamb:
PR #93 still needs to be approved.

Issue #102, opened 2018-07-17
Conversation Stat Bot


lealgamb:
I'm implementing this bot and I wanted to get your guys' feedback on its features. I'm thinking that this bot should be responsible for providing the following:

- Conversation owner
- Conversation creation time
- Privacy status
- number of users
- number of messages
- most/least active users

On that last point, perhaps it can be formatted as a leader board, for both the top 5 and bottom 5 users. What do you guys think?

Also, unlike NemoBot I'm leaning towards not having a map approach for this bot. Instead, on every @ mention, it should just reply with the information listed above since its purpose is very general. Do you guys agree?

yebrahim:
I'm not sure about the last statistic, it doesn't feel very useful to me. The rest I'm +1 on.
A map-style approach would still work, but can't contain the entire logic obviously. You can still use the map to contain the string-formatted messages: `"privacy" -> "This conversation is %s."`

vocatwk:
I like these features. I would rather prefer it answer each one individually though.

PR #103, opened 2018-07-17
Conversation Stat Bot Created

lealgamb:
Created a new bot responsible for conversation statistics. Currently only replies with "Stats loading...". Issue #102 talks about its design and features.

image


jasminechau2:
This looks good!

src/main/java/codeu/controller/ServerStartupListener.java
```diff
@@ -44,6 +45,8 @@ public void contextInitialized(ServletContextEvent sce) {
 
       NemoBot nemoBot = new NemoBot("../../src/main/java/codeu/model/data/NemoBot.txt");
       botController.registerBot(nemoBot.getMentionKey(), nemoBot);
+      ConversationStatBot conversationStatBot = new ConversationStatBot();
+      botController.registerBot(conversationStatBot.getMentionKey(), conversationStatBot);
```
yebrahim: Did we never get around to making `registerBot` take only the Bot instance, and call its `getMentionKey` method within? I can't remember how that discussion went.

src/main/java/codeu/model/data/ConversationStatBot.java
```diff
@@ -0,0 +1,76 @@
...
+
+    return answerMap.get(keyword);
+  }
+
+  /** 
+  * Scans a message to identify keywords in the map. Ignores case.
+  *
+  * @return null if no word is registered.
+  */
+  private String getKeyword(String message) {
```
yebrahim: This seems like a common utility that we could've pushed into a `Bot` base class (instead of the interface). No need for changes now, just pointing out that now we have a reason for a base class. :)

lealgamb: Ah, that makes sense. We could've avoided copy and pasting code.

src/main/java/codeu/controller/ServerStartupListener.java
lealgamb: I can update this. Is this what similar to what you have in mind?

```
botController.registerBot(bot); // In ServerStartUpListener
...
// In BotController
registerBot(Bot bot) {
botMap.put(bot.getMentionKey(), bot);
...
}
```

yebrahim: Yup, exactly!

lealgamb:
This is how the bot replies with the latest changes. 

image

lealgamb:
The bot now replies with the latest changes.
image

jasminechau2:
This looks amazing!

src/main/java/codeu/model/data/ConversationStatBot.java
```diff
@@ -47,11 +86,41 @@ public String getMentionKey() {
...
+    }
+    else if (keyword.equalsIgnoreCase("created")) {
+      return String.format(botAnswer, getCreation(conversation.getCreationTime()));
+    }
+    else if (keyword.equalsIgnoreCase("privacy")) {
+      return String.format(botAnswer, getPrivacy(conversation.isPrivate()));
+    }
+    else if (keyword.equalsIgnoreCase("users")) {
+      int numberOfUsers = getNumberOfUsers(conversation);
+      if (numberOfUsers > 1) {
```
yebrahim: nit: You could shorten these six lines to:
`return String.format(botAnswer, numberOfUsers) + (numberOfUsers > 1 ? "s." : ".");`
Or if you don't like the long line:
```
String suffix = (numberOfUsers > 1 ? "s." : ".");
return String.format(botAnswer, numberOfUsers) + suffix;
```

PR #104, opened 2018-07-17
Can edit members of conversation(Add/remove users)

### LINKED ISSUE #98: Add the ability to remove users from conversations
Hey guys, just wanted to discuss the design for this. Should I add a new modal for this or do it in the same modal as add users?

vocatwk:
At this point, any user can edit the members list by adding/removing any user they want as long as the list has atleast one member. Closing the modal, will discard any changes they didn't save and saving changes will update the list.

I was thinking of restricting this ability to only creators of conversation. what do you guys think?

resolves #98

### DIFF
--- a/src/main/java/codeu/controller/ChatServlet.java
+++ b/src/main/java/codeu/controller/ChatServlet.java
@@ -268,6 +268,7 @@ public class ChatServlet extends HttpServlet {
 
       if(userNameArray == null || userNameArray.length == 0){
         response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+        response.getWriter().write("Conversation must have at least one member.");
         return;
       }
 
--- a/src/main/java/codeu/controller/ChatServlet.java
+++ b/src/main/java/codeu/controller/ChatServlet.java
@@ -115,6 +115,15 @@ public class ChatServlet extends HttpServlet {
       return;
     }
 
+    String purpose = request.getHeader("purpose");
+    if(purpose != null && purpose.equals("Get members")){
+      String json = new Gson().toJson(conversation.getMembers());
+      response.setContentType("application/json");
+      response.setCharacterEncoding("UTF-8");
+      response.getWriter().write(json);
+      return;
+    }
+
     UUID conversationId = conversation.getId();
 
     List<Message> messages = messageStore.getMessagesInConversation(conversationId);
@@ -243,7 +252,7 @@ public class ChatServlet extends HttpServlet {
       }
       conversationStore.updateConversation(conversation);
     }
-    else if(purpose.equals("Adding users")){
+    else if(purpose.equals("Setting users")){
 
       String jsonString = request.getReader().readLine();
       String cleanedJsonString = Jsoup.clean(jsonString, Whitelist.none());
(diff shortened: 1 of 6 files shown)


src/main/java/codeu/controller/ChatServlet.java
```diff
@@ -115,6 +115,15 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
       return;
     }
 
+    String purpose = request.getHeader("purpose");
```
yebrahim: This is fine for now, but generally when your GET handler is serving too many different purposes, it's probably better to break it down to multiple handlers.
Ideally, we'd have a GET handler for each resource. So a handler at `/conversations/<convo_id>/members` would return the list of members. We can then also have a POST handler at the same endpoint to set those members.

No action needed for here though, just discussing a more widely accepted convention. :)

src/main/java/codeu/controller/ChatServlet.java
```diff
@@ -257,21 +266,21 @@ else if(purpose.equals("Adding users")){
         return;
       }
 
+      if(userNameArray == null || userNameArray.length == 0){
+        response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
```
yebrahim: It might be useful to return a suitable error message here along with setting the status. Something like:
`response.getWriter().write("Conversation must have at least one member.");`.
Ideally, the client code would also catch these errors and display them nicely to the user, but it's probably fine not to do that now since we're not really covering all error cases in client code. I'd say the server should still make sure it's sending reasonable error messages.

src/main/webapp/WEB-INF/view/chat.jsp
```diff
@@ -177,8 +234,8 @@ String membersOfConversation = (String) request.getAttribute("membersOfConversat
             <i class="fa fa-cog"> </i>
           </button>
           <div class="dropdown-menu" aria-labelledby="dropdownMenu2">
-            <button id = "privacySettingButton" class="dropdown-item" type="button"><%= privacySettingButtonValue %></button>
-            <button class="dropdown-item btn btn-primary" type="button" data-toggle="modal" data-target="#addUsersModal">Add Users</button>
+            <button id="privacySettingButton" class="dropdown-item" type="button"><%= privacySettingButtonValue %></button>
+            <button id="modalOpenerButton" class="dropdown-item btn btn-primary" type="button" data-toggle="modal" data-target="#addUsersModal">Add Users</button>
```
yebrahim: This button should say "Edit Users" or "Edit Members" instead?

src/main/webapp/WEB-INF/view/addUserBox.jsp
```diff
@@ -21,7 +21,7 @@
       </div>
       <div class="modal-footer">
         <button id="AddUsersModalCloseButton" type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
```
yebrahim: We'll need to change references to "addUsers" or "add-users" everywhere. Maybe "setUsers"?

src/main/java/codeu/controller/ChatServlet.java
vocatwk: yeah that would have made things much more organized.

PR #107, opened 2018-07-19
Fixed deployment problem

lealgamb:
Moved `BotController` declaration into try-catch block. This seemed to stop giving me the error mentioned in the chat.


yebrahim:
Is this reproducible locally at all?
Do you have any guesses as to why this fixes it?

lealgamb:
Everything worked locally on my end. My guess is that putting it out of the try-catch block was calling on Datastore before the server even tried to start? Also, I didn't fully run `mvn appengine:update`. However, prior to this change I got an error at about 8% and after it was going through. I just stopped it sent I was working on master.

lealgamb:
That makes sense. One liners are the worst haha.

PR #109, opened 2018-07-19
Fixed file path

### LINKED ISSUE #108: Update Text File Paths
I'm thinking of placing the text files in `webapp` since it's where the favicon and `index.jsp` are located.

lealgamb:
Fixes #108


src/main/java/codeu/model/data/NemoBot.java
```diff
@@ -22,7 +23,7 @@
   /**
   * Constructs a new Bot.
   */
-  public NemoBot(String filePath) {
+  public NemoBot() {
```
yebrahim: You're removing the file from the constructor here, but passing it up in the listener.

src/main/java/codeu/model/data/NemoBot.java
```diff
@@ -1,6 +1,7 @@
 package codeu.model.data;
 
 import java.io.BufferedReader;
+import java.io.File;
```
yebrahim: Not needed?

src/test/java/codeu/controller/BotControllerTest.java
```diff
@@ -15,7 +15,7 @@
 
   private BotController botController;
 
-  private final String NEMO_FILE_PATH = "./src/main/java/codeu/model/data/NemoBot.txt";
+  private final String NEMO_FILE_PATH = "./src/main/webapp/data/NemoBot.txt";
```
yebrahim: Can't this be built the same way? (using `classLoader.getResource`)

src/main/java/codeu/model/data/NemoBot.java
```diff
@@ -22,7 +23,7 @@
   /**
   * Constructs a new Bot.
   */
-  public NemoBot(String filePath) {
```
yebrahim: Why the changes in this file? We are actually passing the file path not the file right? I'd say just re-checkout this file from master to revert all changes.

lealgamb:
I was able to open the file by moving the code from `ServerStartupListener` to `NemoBot` and using an `InputStream`. Sorry for the previous changes. This works locally and also for tests.

lealgamb:
Should I try it from this branch or merge then deploy from master?

yebrahim:
Yes. Ideally we should have a staging environment (another app in the same project, or an entirely different project) to try things before pushing to production. We don't. :)

lealgamb:
That would be helpful. Just to be sure, that's a yes to try from this branch?

yebrahim:
Yes please go ahead. Worst case scenario is we’ll redeploy from master.

PR #115, opened 2018-07-20
Fixed multiple bot instances

### LINKED ISSUE #113: Multiple Instances of NemoBot
I noticed that each time `ServerStartupListener` there's a new `NemoBot` instance. This results in multiple `NemoBot` users. I tried changing how `BotController` adds bots to the `UserStore`. 

```
private void registerBotAsUser(Bot bot) {
  User botUser = userStore.getUser(bot.getName());
  if (botUser == null) {
    userStore.addUser((User)bot);
  }
  else {
    userStore.updateUser(botUser);
  }
}
```
However, this then breaks the code when displaying messages.
In `chat.jsp`: `String author = UserStore.getInstance().getUser(message.getAuthorId()).getName();`

I think this can be solved by setting a static id, and to follow this maybe we can also set a static creation time.

lealgamb:
To stop the Bot Revolution 🤖 , a bot will only be added to the `UserStore` if it has a unique `UUID` field. The way this is done is by using `getUUIDFromBytes()` by passing in the bytes of the bot's name. The check for unique Id's was added in `UserStore`. A test was also added for this.

Now that the bot classes are completed, I'll send a new PR with tests for `NemoBot` and `ConversationStatBot`, and update the tests for `BotController` if needed.

Fixes #113


src/main/java/codeu/controller/BotController.java
```diff
@@ -49,9 +51,32 @@ void setUserStore(UserStore userStore) {
 
   /** Adds an entry to the map and to the UserStore. */
   public void registerBot(Bot bot) {
-    botMap.put(bot.getMentionKey().toLowerCase(), bot);
+    // Check if the bot is already in the UserStore
```
yebrahim: Are we making sure names are unique while signing up?

lealgamb: We only check this if it's all in the same case. I noticed on production we have "user1" and "User1". Perhaps checking if a username contains "bot", ignoring case, can be a quick fix since we really only care about unique bot names.

yebrahim: What do you mean by 'checking if a username contains "bot"'? You mean rejecting those usernames on sign up, or for detecting whether a bot with this name has been registered?

Here's another thought. How about using a pinned value for the bot UUIDs like you initially did? Maybe we can hash the bot name and use that as a UUID. When starting up, you'd look up any users whose UUID is equal to this bot's name after applying the same hash function, and that is how you can identify bots.

lealgamb: I guess both. My idea is to restrict "bot" as a keyword for actual bots. 

I really like that idea, thanks for suggesting it. So basically if there aren't any matching users with the same UUID, `BotController` can add the bot to the `UserStore`. And if there are, then there would be no need for that.

yebrahim: Yes. It's reasonable that the UUID is used for identifying bots, since it's really meant to be a unique field, unlike name.

lealgamb:
Instead of hashing the Bot's names I got their names' bytes to take advantage of UUID's `getUUIDFromBytes` method.

src/main/java/codeu/controller/BotController.java
```diff
@@ -51,7 +51,10 @@ void setUserStore(UserStore userStore) {
   public void registerBot(Bot bot) {
     botMap.put(bot.getMentionKey().toLowerCase(), bot);
 
-    userStore.addUser((User)bot);
+    // First time adding to the UserStore
```
yebrahim: To me this makes sense to be in the `UserStore` class, since you should never add two users with the same id. Not a hard comment though, so I'm fine if you want to keep it here.

src/main/java/codeu/model/data/ConversationStatBot.java
```diff
@@ -30,7 +29,7 @@
   */
   public ConversationStatBot() {
     super(
-        UUID.randomUUID(), 
+        UUID.nameUUIDFromBytes("ConversationStatBot".getBytes()),
```
yebrahim: One more piece that calls for `Bot` to be a base class instead of an interface. :)

lealgamb: Part of me wants to update this but part of me also wants to work on new features. What do you guys think?

yebrahim: It's not worth it if it's a big change, I thought it was just moving this inside the `addUser` method. If not, never mind. :)

lealgamb: Ah, I did that haha. I was referring to the `Bot` base class instead of the interface.

PR #120, opened 2018-07-30
Updated UI for conversations and chat page

vocatwk:
I updated the UI for conversations page.

I really wanted to implement those icons with the number of messages not yet seen next to each conversation like those in @yebrahim's design cause they just look really cool to have. But I'm not sure if it's that important since the user will see those messages were sent in his notifications page. But I still think it would be a cool feature to have. Let me know what you guys think.

Also while I was thinking of the above, I realized we'd have to account for messages coming in while the user is in the chat. Right now, it won't be difficult to implement because we only update messages in the chat on page reload but that got me thinking that getting those messages without having to refresh might also be good to have. It could be very inefficient at first like using the fetch API to get all the messages every 5000ms and then to improve we can use sockets I think(not really sure about this though). Let me know what you guys think about this too.

I changed the chat page too and updated conversations page a bit. I also added wrap property to the list of edited members.

EDIT #2 

I added back the activityFeed link(I had accidentally taken it out from the navbar). I added a scroll bar to the search results and queued users and added badges to the conversation with the number of unseen messages. I'm working on designing the chat box to look more like messages instead of bullet points. 

screen shot 2018-07-31 at 6 25 36 am
screen shot 2018-08-01 at 6 39 17 pm
screen shot 2018-08-01 at 7 09 03 pm
screen shot 2018-08-01 at 7 35 42 pm
screen shot 2018-08-01 at 7 54 48 pm

### DIFF
--- a/src/main/webapp/WEB-INF/view/conversations.jsp
+++ b/src/main/webapp/WEB-INF/view/conversations.jsp
@@ -24,55 +24,74 @@
 </head>
 <body>
 
-  <div id="container">
+  <div id="conversationContainer">
+    <div id="conversationInput">
 
-    <% if(request.getAttribute("error") != null){ %>
-        <h2 style="color:red"><%= request.getAttribute("error") %></h2>
-    <% } %>
+      <% if(request.getAttribute("error") != null){ %>
+          <h2 style="color:red"><%= request.getAttribute("error") %></h2>
+      <% } %>
 
-    <% if(request.getSession().getAttribute("user") != null){ %>
-      <h1>New Conversation</h1>
       <form action="/conversations" method="POST">
-          <div class="form-group">
-            <label class="form-control-label">Title:</label>
-          <input type="text" name="conversationTitle">
+        <label for="conversationName">New Conversation</label>
+        <div class="form-row align-items-center">
+          <div class="col-auto">
+            <input type="text" class="form-control mb-2" id="conversationName" placeholder="Conversation Title">
+          </div>
+          <div class="col-auto">
+            <button type="submit" class="btn btn-primary mb-2">Create</button>
+          </div>
         </div>
-
-        <button type="submit">Create</button>
       </form>
 
-      <hr/>
-    <% } %>
+    </div>
 
-    <h1>Conversations</h1>
+    <br>
 
-    <%
-    List<Conversation> conversations =
-      (List<Conversation>) request.getAttribute("conversations");
-    if(conversations == null || conversations.isEmpty()){
-    %>
-      <p>Create a conversation to get started.</p>
-    <%
-    }
-    else{
-    %>
-      <ul class="mdl-list">
-    <%
-      for(Conversation conversation : conversations){
-        if(!conversation.containsMember(navBarUsername)){
-          continue;
+    <div id="conversationList">
+
+      <label for="card">Conversations</label>
+
+      <%
+      Boolean listIsEmpty = true;
+      Boolean oneFound = false;
+      List<Conversation> conversations =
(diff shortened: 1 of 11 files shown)


lealgamb:
This looks great! I think that the icons look great, but if it's not a priority we can always return to it if we have some time at the end of the week. 

Since we would only see the number of messages on the conversations page, then I think that getting the number of messages every time the user navigates to this page should work. I might not be thinking exactly of what you have in mind though.

jasminechau2:
I love it the logo is so cool

src/main/webapp/WEB-INF/view/conversations.jsp
```diff
@@ -24,55 +24,74 @@
...
+            <div class="card">
+            <table class="table table-bordered" id="conversationsTable">
+              <tbody>
+      <%
+            listIsEmpty=false;
+            oneFound=true;
+          }
+      %>
+          <tr><td>
+            <a href="/chat/<%= conversation.getId() %>"> <%= conversation.getTitle() %></a></li>
```
yebrahim: What does the `</li>` close?

src/main/webapp/css/main.css
```diff
@@ -62,6 +62,30 @@ body {
...
 
+#conversationContainer{
+  padding-top: 50px;
+  padding-left: 50px;
+  width: 600px;
+}
+
+.card{
+  overflow-x: scroll;
+  max-height: 500px;
```
yebrahim: nit: It's better to have this as a percentage than a fixed value, otherwise we're not making use of larger screens.

yebrahim:
Yes, implementing the badges will be tricky, but it's doable. You can have the server return the number of unread messages along with each conversation object in the GET endpoint for conversations.

You're exactly right about loading incoming messages live. The easiest implementation would use a combination of interval polling and AJAX to update the messages. A more involved implementation would use websockets to keep a two-way connection between each client and the server.

src/main/webapp/WEB-INF/view/conversations.jsp
vocatwk: Oops, forgot to delete.

src/main/webapp/css/main.css
```diff
@@ -152,7 +152,13 @@ body {
 /* chat page messages. */
 
 #messagesContainer {
-    padding: 10px;
+  padding: 10px;
+}
+
+#userResult{
+  max-height: 300px;
+  overflow-y:scroll;
```
yebrahim: Let's make this `overflow-y: auto` instead. `scroll` makes the scrollbar always visible. It's ok if you're doing that on purpose though.

src/main/webapp/css/main.css
```diff
@@ -245,6 +245,8 @@ button {
 #selectedUserList {
   display: flex;
   flex-wrap : wrap;
+  overflow-y: scroll;
```
yebrahim: Same here.

src/main/webapp/css/main.css
yebrahim: This is a more general comment, maybe we should do that all over.

vocatwk: It's visible only when you scroll. It quickly disappears when you stop but I took the screen shot while scrolling just to make it clear in the photos. But if auto is preferred, I can do that.

yebrahim: That's because you're on a mac. Windows and Linux machines will show a fat ugly scrollbar all the time. 🙀 
As I said though, it's fine if this is the intended behavior, just raising the point.

vocatwk: Oh 😅. In that case, yeah I think auto is better. Thanks!

PR #122, opened 2018-07-31
Activity Feed UI

lealgamb:
New look for the Activity Feed. Let me know if you think any more styling is needed.

image

### DIFF
--- a/src/main/webapp/WEB-INF/view/activityfeed.jsp
+++ b/src/main/webapp/WEB-INF/view/activityfeed.jsp
@@ -18,9 +18,9 @@ ConversationStore conversationStore = ConversationStore.getInstance();
 </head>
 <body>
 
-  <div id="container">
-    <h1>Activity Feed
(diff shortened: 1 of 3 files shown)


src/main/webapp/css/main.css
```diff
@@ -125,8 +125,7 @@ button {
 
 /* This is for the Activity Feed Page */
 #activity {
-  background-color: white;
-  height: 500px;
+  max-height: 500px;
```
yebrahim: Let's make this a percentage instead?

src/main/webapp/WEB-INF/view/activityfeed.jsp
```diff
@@ -19,14 +19,14 @@ ConversationStore conversationStore = ConversationStore.getInstance();
 <body>
 
   <div id="container">
-    <h1>Activity
+    <h1>Activity Feed
```
yebrahim: Would be nice to have some `margin-top` for this element, and maybe make it `<h3>`?

PR #123, opened 2018-07-31
Made a page for Bots

lealgamb:
Based on the feedback we got some people recommended a page with information about the bots we offer so here it is. 

image

### DIFF
--- a/src/main/java/codeu/controller/BotServlet.java
+++ b/src/main/java/codeu/controller/BotServlet.java
@@ -1,0 +1,45 @@
+package codeu.controller;
+
+import codeu.model.data.Bot;
+import codeu.controller.BotController;
+import java.io.IOException;
+import java.util.List;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet class responsible for the bots page. */
+public class BotServlet extends HttpServlet {
+
+  /** Controller class that gives control to Bots. */
+  private BotController botController;
+
+  /** Set up state for handling activty feed requests. */
+  @Override
+  public void init() throws ServletException {
+    super.init();
+    setBotController(BotController.getInstance());
+  }
+
+  /**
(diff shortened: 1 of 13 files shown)


src/main/java/codeu/controller/BotServlet.java
```diff
@@ -0,0 +1,45 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet class responsible for the bots page. */
+public class BotServlet extends HttpServlet {
+
+  /** Controller class that gives control to Bots. */
+  private BotController botController;
+
+  /** Set up state for handling activty feed requests. */
```
yebrahim: Doc.

src/main/webapp/WEB-INF/view/bots.jsp
```diff
@@ -0,0 +1,35 @@
...
+<head>
+  <title>Bots</title>
+  <%@ include file="navbar.jsp" %>
+  <link rel="stylesheet" href="/css/main.css">
+</head>
+<body>
+
+  <div id="container">
+    <h1>Bots
+      <a href="" style="float: right">&#8635;</a></h1>
```
yebrahim: What does this accomplish?

src/main/webapp/css/main.css
```diff
@@ -129,11 +129,9 @@ button {
   overflow-y: scroll;
 }
 
-/* This is for the Search Results Page */
-#results {
```
yebrahim: We not using this anymore or removed by mistake?

lealgamb: Not using it anymore. It was from when the search bar redirected to a results page.

src/main/webapp/WEB-INF/view/bots.jsp
lealgamb: It was to update the bots, but I now realize this only happens when the site is updated. I'll remove it.

src/main/java/codeu/controller/BotServlet.java
lealgamb: Oops, sorry about that.

PR #124, opened 2018-08-02
updating about page

jasminechau2:
I added some styling to the about page and I took some of the information from the readme and added it to the about page. When you hover over our names the text changes and shows the descriptions we put.

### DIFF
--- a/src/main/webapp/about.jsp
+++ b/src/main/webapp/about.jsp
@@ -21,61 +21,51 @@
   <link rel="stylesheet" href="/css/main.css">
 </head>
 <body>
-  
-  <div id="container">
+
+  <div id="aboutChat">
     <div
-      style="width:75%; margin-left:auto; margin-right:auto; margin-top: 50px;">
+      style="width:75%; margin-left:auto; margin-right:auto; margin-top: 50px; text-align:left;">
 
-      <h1>About the CodeU Chat App</h1>
-      <p>
-        This is an example chat application designed to be a starting point
-        for your CodeU project team work. Here's some stuff to think about:
-      </p>
+      <h1>About Nemo Chat</h1>
+     This is a chat app created by Team Nemo. It features:
+      <li> a profile pages for users where they can update their About Me's and view their sent messages.</li>
+      <li> private and public conversations.</li>
+      <li> allows you to add/remove users from a conversation.</li>
+      <li> a notification page that allows you to see when messages and conversations are being made that you are member of.</li>
+      <li> an Activity Feed that allows users to see when new users join, create
+  conversations, send messages, and update their About Me's.</li>
+      <li> has a dynamic search feature that allows users to search for other users.</li>
+      <li> a Bot Framework that allows developers to contribute their own Bots. </li>
+      <li> NemoBot, a bot that functions as an assistant.</li>
+      <li> ConversationStatBot, a bot that provides conversation statistics.</li>
 
-      <ul>
-        <li><strong>Algorithms and data structures:</strong> We've made the app
-            and the code as simple as possible. You will have to extend the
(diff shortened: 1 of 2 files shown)


src/main/webapp/about.jsp
```diff
@@ -21,61 +21,50 @@
...
-        working with your team to plan and make changes. Start by updating the
-        homepage and this about page to tell your users more about your team.
-        This page should also be used to describe the features and improvements
-        you've added.
-      </p>
-      
-      <h1> Our team </h1>
+  <div id="aboutChat">
+    <div
+      style="width:75%; margin-left:auto; margin-right:auto; margin-top: 50px; text-align:left;">
```
yebrahim: It's better to put this in the CSS file, but that's just nitpicking.
Btw, you only need these two rules:
```
width: 75%;
margin: 50px auto;
```

src/main/webapp/about.jsp
```diff
@@ -21,61 +21,50 @@
...
+      <li> a Bot Framework that allows developers to contribute their own Bots. </li>
+      <li> NemoBot, a bot that functions as an assistant.</li>
+      <li> ConversationStatBot, a bot that provides conversation statistics.</li>
 
+    </div>
+  </div>
+<br>
+  <div id="bioTitle">
+    <div style="text-align:center;">
+       <h1> Team Nemo </h1>
```
yebrahim: I'd make this an `<h4>`

src/main/webapp/about.jsp
```diff
@@ -21,61 +21,50 @@
...
+       </div>
+  </div>
 
-      <ul>
-    	  <li> <strong> Jasmine Chau:</strong> I'm from California. I'm a
-    		rising Freshman at Middlebury College majoring in Computer Science.
-        I love baking and memes.</li>
-    	</ul>
+  <div class="veronicaBio">
```
yebrahim: You can use the same class for all three bios, and maybe call it "`bio`".

src/main/webapp/css/main.css
```diff
@@ -168,3 +168,69 @@ button.remove-user-button {
   color: white;
   font-weight: lighter;
 }
+
+.jasmineBio {
```
yebrahim: As I mentioned above, you only need one copy of the bio class here, and you can reuse it for all three bios. You also don't need the `.success` classes down here.

src/main/webapp/css/main.css
```diff
@@ -168,3 +168,69 @@ button.remove-user-button {
...
+    text-align: center;
```
yebrahim: I think `text-align: center` is the only thing you need here. Maybe `line-height` too if you don't want them to be very spaced out. I prefer text a little smaller, but that's just my opinion.

src/main/webapp/about.jsp
```diff
@@ -21,61 +21,50 @@
...
-    		and minoring in Statistics.</li>
-    	</ul>
+      <h1>About Nemo Chat</h1>
+     This is a chat app created by Team Nemo. It features:
+      <li> a profile pages for users where they can update their About Me's and view their sent messages.</li>
+      <li> private and public conversations.</li>
+      <li> allows you to add/remove users from a conversation.</li>
+      <li> a notification page that allows you to see when messages and conversations are being made that you are member of.</li>
+      <li> an Activity Feed that allows users to see when new users join, create
+  conversations, send messages, and update their About Me's.</li>
```
lealgamb: nit: spacing

lealgamb:
This looks good! I agree with @yebrahim's comments about the CSS.

yebrahim:
Looks good for merging, you just need to resolve the conflicts.

PR #125, opened 2018-08-03
Added chat bubbles to chat page

vocatwk:
I will put user profile pictures next to messages once I add profile pictures. I will be working on that next and the design for profile pages. Logged in user's messages will be on the right and others will come in on the left.

screen shot 2018-08-02 at 6 33 41 pm


src/main/webapp/WEB-INF/view/chat.jsp
```diff
@@ -274,18 +274,32 @@ String membersOfConversation = (String) request.getAttribute("membersOfConversat
...
-            %>
-          </ul>
+          <%
+            for (Message message : messages) {
+              String author = UserStore.getInstance().getUser(message.getAuthorId()).getName();
+              if(user.equals(author)){
+          %>
+                <div class="messageContainer right">
+                  <div class="box rightBubble blue">
+                    <%= message.getContent() %>
```
lealgamb: nit: store this is a variable.

src/main/webapp/WEB-INF/view/chat.jsp
```diff
@@ -274,18 +274,32 @@ String membersOfConversation = (String) request.getAttribute("membersOfConversat
...
+                  </div>
+                  <img class="dot rightImg" src="/default-user.png">
+                </div>
+          <%
+              }else{
+          %>
+                <div class="messageContainer">
+                  <img class="dot leftImg" src="/default-user.png">
+                  <div class="box leftBubble">
+                    <a class="author" href="/profile/<%=author%>"> <%=author%> </a>
```
lealgamb: maybe we can add a link on the image of the right hand side user

src/main/webapp/css/main.css
```diff
@@ -161,6 +161,83 @@ body {
...
+
+.box {
+  background: #ffffff;
+  color: #007bff;
+  max-width: 30%;
+  margin-left:25px;
+  margin-top:10px;
+  margin-right:25px;
+  padding: 20px;
+  text-align: center;
```
lealgamb: I prefer `text-align: left` but `center` looks good too.

src/main/webapp/css/main.css
```diff
@@ -161,6 +161,83 @@ body {
   margin-top: 10px;
 }
 
+.messageContainer{
+  width:100%;
```
yebrahim: You only need the display flex here, and maybe the bottom margin, although that's barely visible because of the lateral distance.

src/main/webapp/css/main.css
jasminechau2: I agree

src/main/webapp/css/main.css
```diff
@@ -161,6 +161,83 @@ body {
...
+  position:relative;
+  border-radius:10px;
+}
+
+.box.blue{
+  background: #007bff;
+  color: #ffffff;
+}
+
+.leftBubble:before {
```
yebrahim: You can try to share the common rules with the next class, there's quite a few common lines.

jasminechau2:
could it be possible to change the left-hand chat bubbles colors? Like user 3 is grey and user 4 is orange. This isn't needed just a random idea. This is really awesome!

src/main/webapp/css/main.css
```diff
@@ -161,6 +161,83 @@ body {
...
+
+.dot {
+  height: 50px;
+  width: 50px;
+  background-color: #bbb;
+  border-radius: 50%;
+  display: inline-block;
+}
+
+.leftImg{
```
yebrahim: Instead of these two classes, you can just add `margin: 0 5px;` on `.dot`.

yebrahim:
I like @jasminechau2's idea, colors can be useful in a group conversation. :) That's a separate change though.

src/main/webapp/WEB-INF/view/chat.jsp
vocatwk: Not sure what you mean. I don't think I use this anywhere else.

src/main/webapp/css/main.css
vocatwk: You're right, Looks much better.

vocatwk:
@jasminechau2 nice idea, I can try and do that if I have time in the end.

src/main/webapp/WEB-INF/view/chat.jsp
vocatwk: Do you mean clicking on the image could redirect to the logged in user's profile?

src/main/webapp/WEB-INF/view/chat.jsp
lealgamb: I was referring to the message content, but I forgot the message was in an `if-else`. Nvm. 😃

lealgamb:
@vocatwk sorry I didn't comment before merging. Yeah, clicking on the image can redirect to the logged in user.

PR #128, opened 2018-08-03
Notification ui

jasminechau2:
screen shot 2018-08-02 at 11 18 10 pm

if anyone has suggestions I am open to them

### DIFF
--- a/src/main/webapp/WEB-INF/view/notifications.jsp
+++ b/src/main/webapp/WEB-INF/view/notifications.jsp
@@ -16,8 +16,9 @@
   <link rel="stylesheet" href="/css/main.css">
 </head>
 <body>
-  <h4> Back from fishing? This is what you missed:</h4>
+
   <div id="notifications">
+      <p> Back from fishing? This is what you missed:</p>
       <ul>
     <% for (Event notification : notifications) { %>
       <%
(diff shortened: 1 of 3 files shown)


src/main/webapp/WEB-INF/view/notifications.jsp
```diff
@@ -16,7 +16,7 @@
   <link rel="stylesheet" href="/css/main.css">
 </head>
 <body>
-  <p> Hello! This is your notifcation page</p>
+  <h4> Back from fishing? This is what you missed:</h4>
```
yebrahim: Let's make this smaller. `<p>` is actually fine. If you can add a 20px margin to it it'll also make it cleaner.

src/main/webapp/WEB-INF/view/notifications.jsp
```diff
@@ -28,14 +28,24 @@
       %>
       <%if(type.equals("Conversation")){%>
         <li> <strong> <%= Date.from(notification.getCreationTime()) %>: </strong>
-            <%= Username %> opened a conversation!
-            Conversation title <%= notifcationInfo.get(1)%>
+          <div class="alert alert-info alert-dismissible fade show" role="alert">
+            <strong><%= Username %></strong> opened a conversation!
+            Conversation:   <a href="/chat/<%= notifcationInfo.get(2) %>"> <%= notifcationInfo.get(1)%> </a>
+            <button type="button" class="close" data-dismiss="alert" aria-label="Close">
```
yebrahim: You can remove this button, since it'll just hide the alert but not the date title above it.
Alternatively, move the the date from line 30 above the username here, and make it a `div` so they display on its own line. This way clicking the x hides the entire row.

PR #130, opened 2018-08-03
1 Letter or Number for Username

lealgamb:
I was able to register with an empty username which disabled all links to that profile, so I added this check.


jasminechau2:
Good catch!

PR #131, opened 2018-08-03
Navbar collapses

lealgamb:
The navbar now collapses on smaller screen sizes. 
image

However, I ran into a problem with the search bar results. When the first result is clicked, it triggers the dropdown menu instead.
image

### DIFF
--- a/src/main/webapp/WEB-INF/view/navbar.jsp
+++ b/src/main/webapp/WEB-INF/view/navbar.jsp
@@ -20,30 +20,45 @@
   <% String navBarUsername = (String)request.getSession().getAttribute("user"); %>
   <% User navBarUser = UserStore.getInstance().getUser(navBarUsername);%>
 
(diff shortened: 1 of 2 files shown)


vocatwk:
Could we have the drop down positioned somewhere else when it collapses?

lealgamb:
An easy fix would be to move the dropdown menu to the left of the search bar. Or to move it as far left as possible next to the other navbar items. I personally like it on the right side, but this could be an alternative if you guys are okay with this.

vocatwk:
I'm okay with it if it's giving you a lot of trouble.

jasminechau2:
Do what you need to do

lealgamb:
Alright, thanks.

What do you guys prefer? 
image

or
image

jasminechau2:
I like the first one

vocatwk:
same here. I like the first better.

